PR #4, opened 2015-06-09
Stop sorting keys.

betawaffle:
I don't know why this was done originally, but sorting the keys breaks most multi-dimensional metrics. The test cases seem to have been very lucky.

### DIFF
--- a/testdata/rsyncd.json
+++ b/testdata/rsyncd.json
@@ -58,14 +58,14 @@
     "Program": "rsyncd.mtail",
     "Kind": 0,
     "Keys": [
-      "module",
-      "operation"
+      "operation",
+      "module"
     ],
     "LabelValues": [
       {
         "Labels": [
-          "module",
-          "send"
+          "send",
+          "module"
         ],
         "Value": {
           "Value": 2,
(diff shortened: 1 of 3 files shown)


jaqx0r:
The sorting was to make the tests stable :)  I can improve the comparison
in the tests only to avoid the sort.

  Can you show me an example that is breaking so I can test correctness?

betawaffle:
Sure.

- Define a metric with two (or more) dimensions, in my case, the keys were `page` and `edge`.
- Use values that sort differently the keys, in my case, `page` was a number like `1234567` and `edge` was a word like `comments`.

Both the keys and the values were being sorted before being zipped together, causing `edge` to be matched with `1234567` and `page` to be matched with `comments`.

```
counter my_multidimensional_metric by page, edge

/some pattern with an (?P<edge>\w+) and a (?P<page>\d+)/ {
  my_multidimensional_metric[$page][$edge]++
}
```

The above code results in a metric that looks like this:

```
# TYPE my_multidimensional_metric counter
my_multidimensional_metric{edge="1234567",page="comments",prog="example.mtail",instance="example.com"} 1 1433934627
```

jaqx0r:
Ohhh that is a total facepalm.  Thanks!

PR #11, opened 2015-10-12
Add example for apache combined logs

SuperQ:
Add a parser for the commonly used apache "combined"[0] log format.

[0]: http://httpd.apache.org/docs/2.2/mod/mod_log_config.html


jaqx0r:
Great, thanks!

SuperQ:
Yup, we tried writing an apache mod_status exporter, but it's almost useless since it doesn't have an details.  Just global request count and byte count. :disappointed:

PR #14, opened 2015-11-13
remove socket read goroutine; creating garbage, triggering GC and high CPU

ryancox:
I noticed that once statsd or graphite exporters were enabled, CPU utilization would go crazy after a short while. I played around with pprof and noticed high GC calls. The root cause was the go routine to proactively read and discard incoming socket data performing *large* numbers of non-blocking socket reads. This should be unnecessary to keep the writer unblocked.

### DIFF
--- a/exporter/export.go
+++ b/exporter/export.go
@@ -6,7 +6,6 @@
 package exporter
 
 import (
-	"bytes"
 	"errors"
 	"expvar"
 	"flag"
@@ -99,17 +98,6 @@ func (e *Exporter) writeSocketMetrics(c net.Conn, f formatter, exportTotal *expv
 		exportTotal.Add(1)
 		lc := make(chan *metrics.LabelSet)
 		go m.EmitLabelSets(lc)
-		// This goroutine reads any bytes returned from the remote end, to keep
-		// it unblocked.
-		go func() {
-			var buf bytes.Buffer
-			for {
-				_, err := buf.ReadFrom(c)
-				if err == nil {
-					return
-				}
-			}
-		}()
 		for l := range lc {
 			line := f(e.hostname, m, l)
 			n, err := fmt.Fprint(c, line)


jaqx0r:
Wow, great, thanks!

jaqx0r:
I just ran the benchmarks after this pull:

before
BenchmarkRsyncdProgram-4	      50	  24429285 ns/op	  12.80 MB/s
BenchmarkSftpProgram-4  	     300	   6155250 ns/op	3491.71 MB/s
BenchmarkDhcpdProgram-4 	       1	8782405473 ns/op	   0.02 MB/s

after
BenchmarkRsyncdProgram-4	     100	  11614059 ns/op	 206.25 MB/s
BenchmarkSftpProgram-4  	     500	   3871499 ns/op	15361.13 MB/s
BenchmarkDhcpdProgram-4 	       1	7545562716 ns/op	   0.03 MB/s

that seems significant.

jaqx0r:
(op and MB are misleading in these benchmarks, they refer to a full pass of the test data and MB really means lines processed.)

ryancox:
Hmmm I wouldn't have anticipated that. The benchmarks do a one-shot with no pushTargets. The affected code path is only hit when pushTargets is > 0.

```
// StartMetricPush pushes metrics to the configured services each interval.
func (e *Exporter) StartMetricPush() {
	if len(e.pushTargets) > 0 {
```

jaqx0r:
Hmm, there is some other oddity happening, then.  Most recent run now:

BenchmarkRsyncdProgram-4	      50	  23636631 ns/op	  13.23 MB/s
BenchmarkSftpProgram-4  	     200	   7499873 ns/op	1594.46 MB/s
BenchmarkDhcpdProgram-4 	       1	9186606961 ns/op	   0.01 MB/s

jaqx0r:
I was too optimistic there. But these wildy variant benchmarks need some
investigation. Thanks again for the PR.

Issue #15, opened 2016-01-03
Timer metric not supported by Prometheus


dylanmei:
When I add a `timer` metric for [a HAProxy log tailer](https://github.com/dylanmei/haproxy_mtail/blob/master/haproxy.mtail), Prometheus can't scrape the /metrics endpoint.

    text format parsing error in line 13: unknown metric type "timer"

The `timer` metric works well with the StatsD exporter, but to expose this kind of information to the pull-based Prometheus they use [summaries](http://prometheus.io/docs/practices/histograms/) for metrics representing request durations or response sizes. 

Things that could be done
1. Don't publish `timer` metrics via Prometheus export
2. Add an explicit `summary` metric type 
3. Back a `timer` with a `summary` metric, but only expose it fully during Prometheus export

Items 2. and 3. may or may not include bringing in the [Prometheus packages](https://godoc.org/github.com/prometheus/client_golang/prometheus#Summary).

jaqx0r:
Ugh, this is a simple bug... The timer metric exists only to specialise the
output for statsd. It should be treated as a gauge type by every other
export.

dylanmei:
I see, so `timer` is simply a gauge but for StatsD it's formatted differently.

[This kind of change](https://github.com/dylanmei/mtail/commit/1bc2ec1ad7955089434e4071831ca32862ef383d) fixes it for Prometheus by examining the metric `kind` before printing. If you're fine with that I can repeat the solution for other exporters. But ultimately this seems like a design problem.

jaqx0r:
Yes, that's right.

I don't love it, but it was a feature request and I'm happy to support a simple type variance like this if it helps someone.

Your change looks good.  I suppose I'd assumed that this shouldn't have happened, so I will think if there's a way to refactor so that it's less awful, but in the meantime if you'd send that PR I'll take it right away.

PR #16, opened 2016-01-06
Timer as gauge for Prometheus/CollectD

dylanmei:
Related to Issue #15 

The `timer` metric is specific to the StatsD exporter. For the other exporters, it should be exposed as a `gauge`.

Prometheus and CollectD exporters publish their metrics with the type/kind info. In these cases, we ensure that `timer` is published as `gauge`.

Graphite and Varz don't expect metric type/kind info so I've not made any changes there.


dylanmei:
-- also I have completed the CLA.

jaqx0r:
Thanks!

PR #22, opened 2016-04-24
Output string constants when dumping bytecode.

kormat:
This makes it possible to follow the 'str' instructions.


jaqx0r:
Thanks!

Issue #18, opened 2016-04-06
Pattern negation


kormat:
Is there any way to have an action happen if a pattern /doesn't/ match? Go's regex library doesn't support negative-look-ahead, but i can't find any syntax in the mtail language which allows me to work around this.

What i'm trying to do is something like this:
```
/pattern1/ {
  do pattern1 stuff
}
/pattern2/ {
  do pattern2 stuff
}
otherwise {
  do other stuff
}
```
In my case the otherwise bit happens in the absence of the first two. Ideally, being able to chain if/else would be great, but in a pinch a simple 'else' keyword would allow this to be represented as:
```
/pattern1/ {
  do pattern1 stuff
} else {
  /pattern2/ {
    do pattern2 stuff
  } else {
    do other stuff
  }
}
```

jaqx0r:
There's no negative matching, but the else or otherwise clause looks like a
good idea.

kormat:
I've been trying to map out the grammar of the existing language. Here's what i have so far:
https://gist.github.com/kormat/33bfceedecaea2b36effc8eeca650727

jaqx0r:
I'd totally forgotten about this request. I think I can get to it this
weekend. Thanks for the patch!

jaqx0r:
That gist looks pretty accurate.  The transformation from bison grammar
into EBNF should be mechanical.

The only question is, what should the new behaviour look like and act like?

Currently consecutive cond_exprs are executed regardless of their ordering
or prior expression's valyues, i.e. there's no implicit 'else if' on each
line, they're all individual 'if' blcoks.  So should we alter that
behaviour to break out at the first successful one, as if they were
mutually exclusive, or should an 'otherwise' expression match only if no
other expression does?

I think the least surprising behaviour is the 'only if no other', but
that's less elegant to implement than the single-branch.

A more elegant implementation would require letting the user specify where
the 'else if' and 'else' clauses go, but that makes the language more
verbose.

jaqx0r:
I just realised that this is a switch that doesn't terminate on first match
(like Go's own); the 'otherwise' is the default rule.  This becomes much
easier.

kormat:
I think the 'otherwise' idea is probably a bad one :) Here's what i'm currently thinking, which just requires a single `else` keyword (as the `if` is already implicit):

```
/pattern1/ {
  do pattern1 stuff
} else /pattern2/ {
  do pattern2 stuff
} else {
  do other stuff
}
```

kormat:
[...]

PR #23, opened 2016-04-24
Adding 'otherwise' keyword.

kormat:
This is very unready for actual merging, etc. I mostly just want to get some feedback on the approach.


kormat:
#18

kormat:
https://gist.github.com/kormat/c8655c0ec49d2e1ffc29c2c2a1e6c694 has the test program and input that i've been using, and the one_shot_metrics output.

jaqx0r:
Looks good so far!

Please add some more tests in the vm module: lexer_test.go turns text into a token and position stream, parser_test.go validates that the program correctly recognises input (but not the parse tree)  Nice work on compiler_test. vm_test.go tests how each instruction modifies the machine state after a single bytecode execution cycle.

The gist could be turned into an example program test through exampleProgramTests in ex_test.go.

kormat:
`lexer_test`/`parser_test`/`vm_test`/`unparser` all updated. I'll look at `ex_test` now.

kormat:
Hmm, i've run into a few issues:
- Matching on `/^1$/` doesn't work, as the input still contain trailing newlines. This seems like a bug rather than a feature.
- If the input file has a trailing newline, the script is executed one last time with `''` as the input, which causes the outermost `otherwise` to trigger. This seems like a related bug to the first one.
- `ex_test` doesn't like the timestamps in my `.golden`, but i can't see how the existing timestamps are allowed.

kormat:
I've confirmed that the first two issues are actually bugs, and sent #25 to fix them.

kormat:
Found a way to pacify `ex_test`'s passion for timestamps. (This PR now depends on #25 also).

jaqx0r:
Wow, thanks!

examples/otherwise.mtail
```diff
@@ -0,0 +1,18 @@
+counter yes
+counter maybe
+counter no
+
+# To make ex_test.go happy
+strptime("2016-04-25T20:14:42Z", "2006-01-02T15:04:05Z07:00")
```
jaqx0r: neat hack!

Issue #27, opened 2016-04-29
Hangs on sigterm


grobie:
A sigterm doesn't always cleanly terminate mtail. The process just hangs on `23:53:07.163583   25177 mtail.go:255] Shutdown requested.`

jaqx0r:
I can't reproduce this, but this means shutdown has started and something
is blocking and not receiving the shutdown channel notice.

Can you get mtail to dump the goroutine stacks when this happens?

Before running, set GOTRACEBACK=2 environment variable, and then send kill
-QUIT to the process after you send the sigterm and it hangs.

More info here:
https://golang.org/pkg/runtime/

grobie:
https://gist.github.com/grobie/a3e11009b8397cd5ef1adfde655de2a5

jaqx0r:
Goroutines 14 and 15 look like culprits here.

jaqx0r:
Can you tell me what git commit you have built mtail at so I can correlate those stacktraces?

There's a failed watch remove in the logbefore it looks like it tries to add a watch, and blocks on the lock (the semacquire) which appears to block the chan send goroutine.  Just a hypotheis at this stage.

grobie:
It was an earlier commit version, but I can reproduce this on master (6e0eac2) as well. https://gist.github.com/grobie/c2c53a3a565605efa0485114faf460dc

jaqx0r:
Thanks!

jaqx0r:
Yep I think what is happening is tail.go:208 (goroutine 41 in the last gist) is blocking the run() loop in the same routine from taking the next item off the queue, which causes the [chan send] in 40 to block; which means the watcher cannot receive the shutdown message (that Events is closed, from mtail.shutdownHandler()->tailer.Close()->logwatcher->Close())

Why the goroutine 41 is blocked on the semacquire is curious.    fsnotify.v1/inotify.go says that it's waiting for "ignoreLinux" deletes maps, ignoreLinux() only runs in the readevents loop which is running in routine 34, and that one is blocked in a syscall to wait on the fd for new notifications.

I wonder what happened to fsnotify in the logs earlier when:
I0505 01:58:15.582756   29528 tail.go:210] Failed removing watches on/home/grobie/code/go/src/github.com/google/mtail/test/logs/current
occurred.
I'm going to patch the log message to be more verbose, then I'll have to get you to run the test again.

jaqx0r:
[...]

PR #28, opened 2016-05-24
Fix deadlock on removal of watches

beorn7:
fsnotify has the inherent problem that the Watcher.Remove method
relies on events still being read from the events channel
(https://github.com/fsnotify/fsnotify/blob/master/inotify.go#L158). However,
the mtail code stops reading from the event channel until
Watcher.Remove has returned. Certain events, happening with the right
timing, can trigger this deadlock. This leads to
https://github.com/google/mtail/issues/27 , i.e. now and then, on
systems where logfiles are rotated away, mtail hangs.

This PR fixes the problem by running the Watcher.Remove call in its
own goroutine. I'm not entirely sure about undesired side effects, but
in our large-scale mtail setup at SoundCloud, it fixed the original
problem and didn't cause any new ones so far.

Ideally, we should have a test for the bug, but it's hard to recreate
the exact course of events reliably.

And obviously, I think the real problem is how fsnotify handles this
internally.

@jaqx0r

### DIFF
--- a/tailer/tail.go
+++ b/tailer/tail.go
@@ -208,10 +208,13 @@ func (t *Tailer) handleLogCreate(pathname string) {
 			// flush the old log, pathname is still an index into t.files with the old inode.
 			t.handleLogUpdate(pathname)
 			fd.Close()
-			err := t.w.Remove(pathname)
-			if err != nil {
-				glog.Infof("Failed removing watches on %s: %s", pathname, err)
-			}
+			go func() {
+				// Run in goroutine as Remove may block waiting on event processing.
+				err := t.w.Remove(pathname)
+				if err != nil {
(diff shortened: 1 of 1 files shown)


beorn7:
I signed it!
However, the procedure is buggy and doesn't let me specify my real email address, despite registered with my google account. I leave it to you to figure it out.

The test seems flaky, Works locally.

jaqx0r:
Eye roll at the robot. Thanks for the PR!  I was thinking of something
along these lines too.

SuperQ:
Confirmed that this fixes the problem.  ~200 mtail instances watching rotating logs every couple min.  No crashes over the last day, normally would see 15%/day crash.

jaqx0r:
Nice work. Will merge soon!

PR #31, opened 2016-07-04
convert type for other platforms

jarro2783:
This pull request converts the type of an inode in the stat object to the type required by the function. This enables mtail to compile on platforms other than linux and darwin.

### DIFF
--- a/tailer/tail.go
+++ b/tailer/tail.go
@@ -183,7 +183,7 @@ func inode(f os.FileInfo) uint64 {
 	}
 	switch s := s.(type) {
 	case *syscall.Stat_t:
-		return s.Ino
+		return uint64(s.Ino)
 	default:
 		return 0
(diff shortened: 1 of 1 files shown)


jarro2783:
I have signed the CLA.

tailer/tail.go
```diff
@@ -183,7 +183,7 @@ func inode(f os.FileInfo) uint64 {
 	}
 	switch s := s.(type) {
 	case *syscall.Stat_t:
-		return s.Ino
+		return uint64(s.Ino)
```
jaqx0r: I'm curious, what is a syscall.Stat_t.Ino on your platform?  (What platform?)

jaqx0r: Ah I see on freebsd it's a uint32.  Looks good to me.

PR #39, opened 2016-07-28
Replace glog.Error with glog.Errorf

themalkolm:
Due having a placeholder in the string by bet that this is a typo.


themalkolm:
I signed it!

jaqx0r:
Thanks!

PR #42, opened 2016-09-18
Return error when program loading failed

skyan:
I just run mtail, test a simple program, and found that when program has syntax error, mtail will not print any error messages. 

So I hack the code and fix this problem.

### DIFF
--- a/vm/loader.go
+++ b/vm/loader.go
@@ -64,7 +64,10 @@ func (l *Loader) LoadProgs(programPath string) error {
 			if fi.IsDir() {
 				continue
 			}
-			l.LoadProg(path.Join(programPath, fi.Name()))
+			err = l.LoadProg(path.Join(programPath, fi.Name()))
+			if err != nil {
+				return err
+			}
 		}
 		return nil
 	default:


jaqx0r:
The system log should show all compile errors at runtime. There is also a
pair of flags to compile only "compile_only" and print errors and to run a
test over sample logs "oneshot". These flags will exit with an error but
the program will not exit otherwise in case it is running other programs at
the same time.

vm/loader.go
```diff
@@ -64,7 +64,10 @@ func (l *Loader) LoadProgs(programPath string) error {
 			if fi.IsDir() {
 				continue
 			}
-			l.LoadProg(path.Join(programPath, fi.Name()))
+			err = l.LoadProg(path.Join(programPath, fi.Name()))
+			if err != nil {
+				return err
```
jaqx0r: this should capture the error in a slice of errors, and return them all after the loop is complete.

jaqx0r:
I see what you mean now.  it's a bug the directory loader didn't return any error at all.  Looks good.

skyan:
That's nice, thanks for review.

PR #45, opened 2016-10-11
exporter: add prefix option for graphite/statsd metrics

filippog:
Useful for example to further namespace metrics by host or other criteria.

### DIFF
--- a/exporter/collectd.go
+++ b/exporter/collectd.go
@@ -13,12 +13,14 @@ import (
 )
 
 const (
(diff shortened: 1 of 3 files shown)


filippog:
I signed it!

filippog:
@jaqx0r added collectd prefix to `plugin` collectd name, AFAICT the best fit, what do you think?
re: flags testing, I'd be happy to add that, where would it be the best place?

jaqx0r:
The change looks good. The test should go into export_test and set the flag to a nonnull string.

PR #46, opened 2016-10-12
Add response histograms to apache_metrics example.

SuperQ:
Sorry, microseconds is the only reasonably compatible time until mtail supports floats metrics.

CC @beorn7


examples/apache_metrics.mtail
```diff
@@ -3,11 +3,12 @@
 
 # Parser for a metrics-friendly apache log format
 # LogFormat "%v:%p %R %m %>s %H conn=%X %D %O %I %k" metrics
-counter apache_http_requests_total by server_port, handler, request_method, request_status, request_protocol
 counter apache_http_connections_total by server_port, handler, request_method, request_status, request_protocol, connection_status
-counter apache_http_requests_time_microseconds_total by server_port, handler, request_method, request_status, request_protocol
-counter apache_http_sent_bytes_total by server_port, handler, request_method, request_status, request_protocol
 counter apache_http_received_bytes_total by server_port, handler, request_method, request_status, request_protocol
+counter apache_http_request_time_microseconds by quantile, server_port, handler, request_method, request_status, request_protocol
```
beorn7: "quantile" doesn't seem like the right term here. It's the boundary of the latency bucket we are talking about here. So "bucket" or "upper_bound" or something?

beorn7:
10x buckets are really coarse. Check out our default buckets for a "network service": https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go#L54-L59

Even those are pretty coarse, and only meant as a starting point before you manually tailor the buckets and zoom into the area you are interested in.

examples/apache_metrics.mtail
SuperQ: I was just using the labeling used by other similar metrics.  This should result in normal Prometheus-style cumulative histogram labeling.

SuperQ:
This is an example file, I didn't want to get overly granular, but just provided a starting point example for "normal" apache servers.  But I can adjust them.  Part of the problem is apache servers sometimes also handle very long-lived requests.

SuperQ:
@beorn7 `quantile` in this case is the label used in the output of the text metrics.  I have updated this to `le`.

beorn7:
OK, so the labels should be fine now, but it would be good to see an example output.

WRT bucket scheme: Especially because it is an example and because Apache may server very long running queries, I'd use the same bucketing as in https://github.com/prometheus/client_golang/blob/master/prometheus/histogram.go#L54-L59 . The buckets there were laid out under the exact same premises.

Some monitoring systems use 4x buckets, and that's already mostly useless. 10x is really not a good role model, so it shouldn't be used in an example.

SuperQ:
@beorn7 Ok, I've updated the buckets to follow the recommended defaults.

beorn7:
Thanks.

Could you attach a sample output to this PR so that we can check the final outcome?

jaqx0r:
Thanks!

PR #47, opened 2016-10-26
Fix mtail target in Makefile

xbglowx:


jaqx0r:
The commit summary is incorrect, "Fix" should say "Rename".  Why do you
want to change the install target to be 'mtail' ?

xbglowx:
Without this fix/rename, building from source wouldn't work for me. It would say:

    make: Nothing to be done for `all'.

Maybe I am missing something, but I couldn't find anything in the build docs.

xbglowx:
I believe my Go environment is setup correctly:

    10041 ➜  ~  cd $GOPATH/src/github.com/google/mtail
    10042 ➜  ~/go/src/github.com/google/mtail git:(master) git pull
    Current branch master is up to date.
    10043 ➜  ~/go/src/github.com/google/mtail git:(master) make
    make: Nothing to be done for `all'.

filippog:
I get the same with GNU make 3.81, my guess is because of `all: mtail` but mtail is a directory and no other rules on how to build it.

jaqx0r:
OK, thanks for the explanation.  Can you make the target list both 'install
mtail' for backwards compat?  I believe the correct syntax is "install
mtail: ..."

xbglowx:
@jaqx0r done.

jaqx0r:
Thanks!

Issue #68, opened 2017-06-30
Numerical expressions / numerical types issues


mjf:
There _may_ be an issue with the numerical expressions or the `mtail`'s type system as such:

    $ cat progs/test.mtail
    counter t by le
    counter t_sum
    
    /^(?P<v>\d+(\.\d+)?)/ {
        $v < 0.5 {
                t["0.5"]++
        }
        $v < 1 {
                t["1"]++
        }
        t["inf"]++
        t_sum += $v
    }

    $ touch test.log
    $ mtail -progs progs -logs test.log

    $ curl -s localhost:3903/metrics
    # TYPE t counter
    # TYPE t_sum counter
    t_sum{prog="test.mtail",instance="test"} 0

    $ echo 0.1 >>test.log
    $ echo 1 >>test.log
    $ echo 1.765 >>test.log

    $ curl -s localhost:3903/metrics
    # TYPE t counter
    # TYPE t_sum counter
    t_sum{prog="test.mtail",instance="test"} 0

Q1: Am I missing something? (BTW: `mtail` itself produces neither warning nor error and simply does nothing in such a configuration.)

Q2: How do I cope with mixed integer and floting-point values in logs (such as the `delay=` and `delays=` values in `postfix` logs)?

**Sidenote:**

I would like to suggest `mtail` behaves similarly to `awk`. Please refer to for instance `gawk(1)` and look up the "Variable Typing And Conversion" paragraph to know more on this topic. The `awk`'s way **is** good-enough (and smart-enough) and would provide much more power to `mtail` users as well as _may_ be compatible with most existing `mtail` programs (not sure).

jaqx0r:
Use --logtostderr to see the compiler logs on standard error.  Raise the
log level with --v=2.

This is a nice test case, thankyou!

The t counter is not being incremented so either the pattern is not
matching, or the comparison in $v < X is not working.  I suspect the latter
because I know I haven't got a unit test for that yet.

I will read the gawk manual, but my uninformed opinion is yes, the language
already copies awk so type conversion should too.

mjf:
@jaqx0r 

I also think the `$v < X` expressions and similar are not working for some reason.

**Off-Topic:**

Does `$v == "literal"` work? It seems that string comparisons does not work at all. I would like to (for example):

    counter foo

    /(?P<w>\w+)/ {
        $w == "foo" {
            foo++
        }
    }

Does it make sense? ... But that does not seems to me to work either atm.

[...]

PR #72, opened 2017-07-07
[WIP] support float64 and string in cmp

hephex:
Comparing two floats doesn't work because only integers are supported in `cmp`, and I'm not able to have a condition like this one in my mtail program:
```
/(?P<time>\d+\.\d+)r/ {
	$time < 0.1 {
		bucket++
	}
}
```
I had a look at #68, and I also added support for strings.

Do you think that the cast rules are correct?


jaqx0r:
This is really good, thanks for doing it!

I was going to ask that you put the compare functions in a new file, compare.go, but I am not sure that's necessary as there's a lot of noise in this directory.  Keep them where they are for now.

I have plans for a formal type cast and some vm support for it to avoid the type switching we have here and in other places (e.g. PopInt) but as that's not here yet, this is great.

Thanks for the contribution!

vm/vm_test.go
```diff
@@ -113,6 +113,41 @@ var instructions = []struct {
 		[]interface{}{2, 2},
 		[]interface{}{},
 		thread{pc: 0, match: false, matches: map[int][]string{}}},
+	{"cmp gt float float",
+		instr{cmp, 1},
+		[]*regexp.Regexp{},
+		[]string{},
+		[]interface{}{"2.0", "1.0"},
```
jaqx0r: Can you convert these tests from strings to floats, and ints later on?  I think we're tripping over an accidental cast from string here.  The previous test case uses ints.

hephex:
Thank you.

Do you already have an idea how to design and implement the formal type cast?

I tried to change the value of some `match` in the tests, but they didn't fail as I expected. Can you check if it happens to you as well?

vm/vm_test.go
jaqx0r: Yep, thanks!

jaqx0r:
Drop [WIP] from the description and I'll merge this as it is.

jaqx0r:
I want to add vm instructions to do the type casting, e.g. (i2s, i2f, and
so on), and either have the type checker add new nodes to the AST to make
the cast explicit, or just have the code generator infer the cast and emit
the instruction.  (either way, the code gen phase needs to emit the
instruction.)

Not failing the test is not good.  Is the change you mean here in this PR?

hephex:
I didn't commit that change, but I managed to reproduce it on master too.

Changing the expected value for `match` in the test doesn't make it fail.

i.e. after this change the test still pass
```
-               thread{pc: 0, match: true, matches: map[int][]string{}}},
+               thread{pc: 0, match: false, matches: map[int][]string{}}},
```

Am I missing anything about how `TestInstrs` works?

jaqx0r:
It really looks like the test is failing to compare the thread state correctly.

hephex:
I added this code to `vm_test.go`, and if I change the thread in a test, it fails now, but `deep.Equal` seems not working in this case.

```
                if diff := deep.Equal(v.t, &tc.expectedThread); diff != nil {
                        t.Errorf("%s: unexpected virtual machine thread state.\n%s", tc.name, diff)
                }
+               if !reflect.DeepEqual(v.t, &tc.expectedThread) {
+                       t.Errorf("%s: unexpected virtual machine thread state.\n", tc.name)
+                       t.Errorf("\t%v", *v.t)
+                       t.Errorf("\t%v", tc.expectedThread)
+               }
```

btw I guess this discussion is for another issue (or PR).

hephex:
`go-test` compares only the exported fields by default, there's a flag to compare the unexported ones: `deep.CompareUnexportedFields`.

https://github.com/go-test/deep/blob/master/deep.go#L171

jaqx0r:
Wow!  Thanks, great find!

I couldn't get the deep.CompareUnexportedFields flag to work straight away, and there's no examples out in the wild, so I used your reflect.DeepEqual code snippet instead.  The tests now fail correctly; this is pushed to master.

PR #75, opened 2017-07-13
Add GitLab-CI yaml

SuperQ:
Add a GitLab CI configuration for building artifacts on GitLab.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,28 @@
+image: golang:1.8
+
+stages:
+  - test
+  - build
+
+before_script:
+  - mkdir ${CI_PROJECT_DIR}/build
+  - mkdir -p ${GOPATH}/src/github.com/google/
+  - ln -s $(pwd) ${GOPATH}/src/github.com/google/mtail
+  - cd ${GOPATH}/src/github.com/google/mtail
+
+test:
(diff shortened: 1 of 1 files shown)


jaqx0r:
Can you elaborate please, what does adding gitlab do and how do we use it?
Assume I am a total noob.

SuperQ:
This allows any gitlab users to have an automatic binary build pipeline.

.gitlab-ci.yml
```diff
@@ -0,0 +1,28 @@
...
+  stage: test
+  allow_failure: true
+  script:
+    - make test
+
+build:
+  stage: build
+  script:
+    - make install
+    - go build -o ${CI_PROJECT_DIR}/build/mtail
```
jaqx0r: I don't understand how this used, why not copy from $GOPATH/bin/mtail ?  Is this really masking a bug in the install?

SuperQ: Hrmm, good question, I didn't actually write this.  I just copied what we had [upstream](https://gitlab.com/gl-infra/mtail/blob/master/.gitlab-ci.yml).

I will look into why this is.

SuperQ:
Tested the config upstream with a `cp` instead of a build.  Works perfectly.

PR #76, opened 2017-07-13
Ignore hidden program files

SuperQ:
Ignore "hidden" program files that start with a `.`.  This avoids
attempting to load temp files deployed by config management systems.

### DIFF
--- a/vm/loader_test.go
+++ b/vm/loader_test.go
@@ -174,3 +174,29 @@ func TestProcessEvents(t *testing.T) {
 
 	}
 }
+
+var testProgFiles = []string{
+	"test.wrongext",
+	"test.mtail",
+	".test",
+}
+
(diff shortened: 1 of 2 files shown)


vm/loader_test.go
```diff
@@ -174,3 +174,29 @@ func TestProcessEvents(t *testing.T) {
...
 	}
 }
+
+var testProgFiles = []string{
+	"test.wrongext",
+	"test.mtail",
+	".test",
+}
+
+func TestLoadProg(t *testing.T) {
```
jaqx0r: You are a legend for adding a test.

SuperQ: I was trying to figure out how to check the contents of type `Loader` to validate that it's not loading the two non-conforming filenames.  But it doesn't look like the struct knows.

PR #78, opened 2017-07-17
[WIP] add support for histogram

hephex:
It still needs some work, testing, and docs, but I'd like your feedback. I'm completely sure about the new datum `Buckets` and the syntax.

A histogram is defined as `histogram <name> with 0, .1, .3, 1.`. It has the same syntax as the other metrics plus the keyword `with` to define the metric's buckets.

A histogram always uses the datum `Buckets`, when a new value
is assigned to the metric (`iset`), it will:

+ increment the counter for the observed value.
+ increment the total count.
+ update the total sum of all the observations.

Thank you.

### DIFF
--- a/emgen/emgen.go
+++ b/emgen/emgen.go
@@ -44,7 +44,7 @@ var table = map[string]node{
 	"decl":          {[][]string{{"hide_spec", "type_spec", "declarator"}}, ""},
 	"hide_spec":     {[][]string{{""}, {"hidden"}}, ""},
 	"declarator":    {[][]string{{"declarator", "by_spec"}, {"declarator", "as_spec"}, {"ID"}, {"STRING"}}, ""},
-	"type_spec":     {[][]string{{"counter"}, {"gauge"}}, ""},
+	"type_spec":     {[][]string{{"counter"}, {"gauge"}, {"histogram"}}, ""},
 	"by_spec":       {[][]string{{"by", "by_expr_list"}}, ""},
 	"by_expr_list":  {[][]string{{"ID"}, {"STRING"}, {"by_expr_list", ",", "ID"}, {"by_expr_list", ",", "STRING"}}, ""},
 	"as_spec":       {[][]string{{"as", "STRING"}}, ""},
--- a/exporter/prometheus.go
+++ b/exporter/prometheus.go
@@ -8,9 +8,11 @@ import (
 	"fmt"
 	"net/http"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/google/mtail/metrics"
+	"github.com/google/mtail/metrics/datum"
 )
 
 var (
@@ -58,6 +60,10 @@ func (e *Exporter) HandlePrometheusMetrics(w http.ResponseWriter, r *http.Reques
 	}
 }
 
+func prometheusMetricLine(name, value string, labels ...string) string {
+	return fmt.Sprintf(prometheusFormat, noHyphens(name), strings.Join(labels, ","), value)
+}
+
 func metricToPrometheus(hostname string, m *metrics.Metric, l *metrics.LabelSet) string {
 	var s []string
 	for k, v := range l.Labels {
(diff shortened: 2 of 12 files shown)


SuperQ:
This is a great idea, it would eliminate the need for my manual histogram hacks. 👍

jaqx0r:
I am wondering aoubt the bucket specification (with_spec).  Rather than have a fixed list of bucket boundaries, we could also generate them, e.g. "I'd like a bucketer that starts at zero, then 1, then the bucket width grows by 1.6x, and repeats 10 times, and inf is the final bucket"

I think we'd need two syntaxes for this.

histogram foo with boundaries 0, 0.1, 0.5, inf
histogram foo with bucketer 0, 1, 1.6, 10

Also could the form re-use "by" instead of adding "with" ?  I feel if you're using prometheus especially, the "with" keyword has different enough meaning between here an promql that it might confuse a user.

histogram foo by 0 0.1 0.5
histogram foo by (0, 1, 1.6,  10)

I think getting the syntax right is important, so if you'd like to break this PR into first the metrics,exporter part and a second one for the compiler (vm/...) part, that would help get merged.  (I can even see you breaking up the cmopiler PR into two, leaving the parser+lexer until last as that's where the syntax bikeshed will happen.)

SuperQ:
In the Prometheus libraries, we generally have 3 options.
* Manual buckets
* Helper function - Linear buckets
* Helper function - Exponential buckets

Using `by` is probably the wrong syntax, as we should support additional labels for histograms.

histogram http_request_duration_seconds buckets 0.01, 0.1, 1.0 by code

Notes:
* You never want a 0 bucket, as it's basically impossible.
* The infinite buckets should be automatic for all histograms as a catch-all as buckets are `<=`.

jaqx0r:
I like that syntax.  "Bucket" makes it clear. (I think singular bucket
rather than buckets so it reads like a verb.)

hephex:
> I think getting the syntax right is important, so if you'd like to break this PR into first the metrics, exporter part and a second one for the compiler (vm/...) part, that would help get merged.

It makes sense; I'll split this into 2/3 PRs.

`bucket` is definitely better than `with` and more explicit. Should the buckets *helper* be a built-in?
e.g.
```
# Manual buckets
histogram http_request_duration_seconds bucket 0.01, 0.1, 1.0 by code
# Linear buckets values: start, width, count
histogram http_request_duration_seconds bucket linear(1.0, 1.0, 5) by code`
# Exponential buckets values: start, factor, count
histogram http_request_duration_seconds bucket exponential(1.0, 1.5, 5) by code`
```

It'd be easy to extend if you need a different type of buckets generator.

SuperQ:
@hephex The helper function idea is nice.

I don't understand the reason to make `buckets` singular, when there are multiple buckets by definition.

jaqx0r:
OK, stickwith 'buckets'

I don't know about the helpers yet, let's leave that until a future PR.

jaqx0r:
Followup to explain: A builtin won't work here (without extra changes, which I won't accept :smile: ) because the declaration doesn't create any bytecode to execute in order to construct a new metric.  Builtins are all custom opcodes in the VM.

I think we require a special syntax in the declaration to indicate which bucketing technique to use; and that we shouldn't consider the helper methods as special like Prometheus does, instead that there are three methods and one of them is explicitly naming them.

mjf:
Hello @jaqx0r and @hephex,

I really like your effort to add the histogram support, I must say! You are definitely going to make people happier with *mtail* (especially those who have recently written some buckets manually in *mtail* code).

Now, for the syntax - what about (in *ABNF*):

    "histogram" Metric "buckets" ["manual" / "linear" / "exponential"] Arguments ["by" Indices]

For the examples given by @hephex it would look like:

    # Manual buckets
    histogram http_request_duration_seconds buckets 0.01, 0.1, 1.0 by code
    # and (for consistency)
    histogram http_request_duration_seconds buckets manual 0.01, 0.1, 1.0 by code

    # Linear buckets values: start, width, count
    histogram http_request_duration_seconds buckets linear 1.0, 1.0, 5  by code

    # Exponential buckets values: start, factor, count
    histogram http_request_duration_seconds buckets exponential 1.0, 1.5, 5 by code

Looks pretty straightforward to me. Although *mtail* is related to the *Prometheus* there is simply no need to make it 100% look and feel like *Prometheus*, right?

SuperQ:
No, the syntax for the mtail config does not need to look like Prometheus.

jaqx0r:
I quite like that syntax, thanks for the suggestion!

Can we assume that histograms must be counters in each bucket?  Or are there cases where they might be gauge types?  (Only because we don't say "histogram counter" now.)

SuperQ:
The individual components (buckets, sum, count) of Histogram metrics are counter metrics by definition.  We're taking individual events and aggregating them into accumulators.  "histogram counter" would be redundant.

jaqx0r:
@SuperQ ok great, that matches my assumption

curlup:
I think it would be very nice to have an option whether histogram counters should be 
* cumulative with only `le` label prom style (as in an example in docs)
* or `lower,upper` bounds label OpenTSDB style with non-cumulative values

PR #83, opened 2017-07-24
Resolves #82 allow specific bind interface

### LINKED ISSUE #82: option should be added to allow explicitly setting interface on which to bind
the singe `-port` option is implemented in such a way as to make it impossible to selectively bind a specific interface.

It takes a string as an argument, but then that string is passed to `http.ListenAndServe` with a bare colon prepended to cause a default `0.0.0.0` binding. 

https://github.com/google/mtail/blob/master/mtail/mtail.go#L288

I propose that either the code be modified to allow for a full `address:port` being passed from the `-port` flag OR adding an additional `interface/address` flag.

The use case is for exporting metrics on a multihomed machine where it is undesirable to have an http listener on every interface.

drawks:
* Add a bit of logic between parsing flags and instantiating mtail
  instance to decide whether to use an implicit global bind
* Updated `MtailServer` to not munge `port` since the value in `o` is
  already corrected.

### DIFF
--- a/main.go
+++ b/main.go
@@ -99,6 +99,10 @@ func main() {
 			glog.Exitf("No logs specified to tail; use -logs or -logfds")
 		}
 	}
+	if _, err := strconv.Atoi(*port); err == nil {
+		// If supplied port is a bare int try prepending ":" for implicit global bind
+		*port = ":" + *port
+	}
 	o := mtail.Options{
 		Progs:                *progs,
 		LogPathPatterns:      logs,
--- a/main.go
+++ b/main.go
@@ -51,8 +51,9 @@ var logs seqStringFlag
 var logFds seqIntFlag
 
 var (
-	port  = flag.String("port", "3903", "HTTP port to listen on.")
-	progs = flag.String("progs", "", "Name of the directory containing mtail programs")
+	port    = flag.Int("port", 3903, "HTTP port to listen on.")
+	address = flag.String("address", "", "IP address on which to bind HTTP listener")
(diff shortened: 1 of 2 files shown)


drawks:
@jaqx0r any chance I can get a merge here? It is a very simple change and preserves backwards compatible behavior. Coverage and test unchanged

mattbostock:
@jaqx0r: What should the new flag be called? Usually one might use `addr`, but it seems a lot of the other flags use `hostport`.

drawks:
It doesn't mean "more than just the listening port", it just allows for a more specific selection of the port to bind. `127.0.0.1:3900` is a port just like `3900` is a port, the existing behavior just makes the address portion of the specification implicit while my PR allows for it to be explicitly set.

jaqx0r:
@drawks It does change the meaning of the flag if you add the bind address into the accepted values -- the flag name is "port" and the flag description says "HTTP port to listen on."

If you proceed with the code as written, I would like the flag renamed to "hostport" or "bind_addr" or something, like @mattbostock suggests.

But I also don't want to sacrifice backwards compatibility, so that why I suggested you add a flag for "address", that you can concatenate with port as you have done on main.go:104 in your branch.  I think this choice is simpler (you don't have to check for a ":" in the port string) and preserves backwards compat.

Thanks.

drawks:
I disagree with your assessment, but I'll make the change as requested, because arguing over the detail seems not a good use of energy.

jaqx0r:
Thanks.

Issue #63, opened 2017-06-08
Support for copytruncate


woojae:
Support for copytruncate when log rotating would be nice.  

copytruncate
              Truncate  the  original log file in place after creating a copy,
              instead of moving the old log file and optionally creating a new
              one.   It  can be used when some program cannot be told to close
              its logfile and thus might continue writing (appending)  to  the
              previous log file forever.  Note that there is a very small time
              slice between copying the file and truncating it, so  some  log-
              ging  data  might be lost.  When this option is used, the create
              option will have no effect, as the old log file stays in  place.

jaqx0r:
This is a weird behaviour.  If the inode hasn't changed then it will be difficult to know if log rotation has occurred.

I think seeing that the seek pointer is well past the size of the file could be a signal.  It sounds like it is not very robust though.

So that I can learn more about it, what log program's manual did you copy this text from?

woojae:
The text was copied from the man of logrotate. 

[https://linux.die.net/man/8/logrotate](https://linux.die.net/man/8/logrotate)

rjmcnamara:
I'd like to see a workaround for this as well.  How often would mtail be able to do the comparison that you suggest (current pointer vs. filesize) ?   What you suggest should be sufficient unless log writes are out of control and rotations are being triggered more quickly than mtail can do the comparison.

jaqx0r:
@woojae, thanks.

I have to check my thinking but as soon as the truncate happens, mtail
should receive a fsnotify event about the file modification; normally this
leads to us trying to read from the last file seek pointer.  At that moment
we can check if that's a valid offset in the file, and reset to zero if it
is not.

The complication is that the dance between fsnotify and file open/seek is
tricky already and I don't have the state in my head as I write this to
reason about it :-)

Sounds like a totally reasonable feature to add though.

Thanks for the report!

Savemech:
Would be also nice, if mtail able to handle HUP signals to start re-read log files, like mtail was restarted, but with compiled mtail.progs
kill -USR1 gives nothing
kill -1 mtail gracefully shutdown 
kill -HUP mtail gracefully shutdown
but would be cool if start reading logs over again
So i enforced to call mtail from my logrotate:
```
/path/to/some/java/app/logs/*log {
        copytruncate
        rotate 7
        missingok
        size 1
        create 644 app app
        compress
postrotate
		pidof mtail || there goes full mtail init command
endscript
}
```
Would be awesome if my logrotate.d/* behold none of this

[...]

PR #101, opened 2017-10-09
Support copy truncate 

ncabatoff:
This is a fix for #63.  Really it's only the last 3 commits that are relevant; I made some changes to this fork last year that I foolishly applied to master, and the only safe way I found to clean that up was to revert them.


ncabatoff:
I forgot to run make testrace.  Seems my new test has a data race, but I don't see how I can fix it with afero memmap files and fakewatcher.  Should I switch the test to work with real files?  I don't mind doing that but it seems inconsistent with the other tests.  Any other ideas?

jaqx0r:
Thanks!

Issue #103, opened 2017-10-11
Add support for =~ (alternatively just ~) and !~ operators


mjf:
Hello,

it would be really nice if the `=~` (alternatively just `~` for compatibility with `awk(1)`, the `=~` is for compatibility with *PromQL* most people are used to here) and `!~` operators vere supported (may need some support for contatenation as well). Examples (probably not all of possible combinations):

    # string
    $foo ~ "bar" {
        # ...
    }

    # regex
    $foo !~ /bar/ {
        # ...
    }

    # const
    const BAR /bar/

    $foo =~ BAR {
        # ...
    }

    # var (obscure but why not)
    /(?P<bar>bar)/ {
        $foo ~ $bar {
            # ...
        }
    }

I would also suggest the `=~` (`~`) and `!~` operators automatically converted *r-values* to type *regex*, if there's any (not sure) otherwise some of the above exmaples would not work. If not acceptable at least manual "casting" should be allowed for *non-regex* or *unknown* types (as it already works for simple matching). Examples:

    $foo ~ // + "bar"
    $foo ~ // + BAR
    $foo ~ // + $bar

Whether to allow "casting" of numerical types is a question.

-----

**Real-world use case**

I use `mtail(1)` to count misc. metrices from mail server log files. I would like to count

1. total count of relays and
2. count of IPv6 relays,

so that I knew how many percent of relays are on IPv6. Example (simplified):

    /,\s+relay=[^[]+\[(?P<relay_ip>[^]]+)\]:/ {
        $relay_ip ~ /:/ {
            counter ipv6_relays_count
            ipv6_relays_count++
        }
    }

Of course workaround exists but it's very ugly. Having the `~` and `!~` operators would make code
maintainable and simpler.

jaqx0r:
This is a good idea. I tried to do something similar with conditional
expressions last week, e.g. /(?P<var>.*)/ && $var == "foo" but the parser
doesn't allow that yet, as I have an error in the way regular expressions
are used as conditionals.  Fixing both will make this work.

Thanks for the report!

mjf:
@jaqx0r For this and I am looking forward to it. (I also hope you could raise priority for this. `mtail(1)` is evolving and this is kind of language design pattern...) Thanks for you great work too!

Kelvin-Chu:
+1 on this.  Having this would be extremely useful.

jaqx0r:
There's some more code required to make this work but I'm whittling it down.

jaqx0r:
[...]

Issue #88, opened 2017-08-01
Allow matching regexes against variables


TheTincho:
Hi,

While writing the Postfix rules (#87), something that would have helped coding and making the rules more solid is to allow to match regexes against variables.

For example, the syslog decorator leaves all interesting text in a $description variable, but then all regexes forcefully have to match against the whole line. Having a perl like binding operator ($var ~ /regex/) would solve this.

Thanks.

TheTincho:
When I said variables, I meant variables and/or capture groups.

jaqx0r:
This came up in issue #36 as well.  I don't have a good answer to that yet,
the match implicitly assumes its on the current line input only.

I like the perl option; its $_ is a lot like the implicit line in mtail.
Thanks for the suggestion!

jaqx0r:
I've tested this in the latest commit and I think it's working as described now.

PR #104, opened 2017-10-11
Add concat function

### LINKED ISSUE #102: Add support for arbitrary string processing of matched fields
Dear all,
I'm using latest mtail (https://github.com/google/mtail/releases/tag/v3.0.0-rc4) for legacy applications prometheus.io style metric reporting.

I believe mtail is good tool but there are plenty of areas to perform better.

I currently want to combine two matched logfile line fields concatenate them and use product as metric's label `errors_total{type=...}`. Is there a way how to achieve it?

Let's look at my mtail rule snippets:
 * option A (preferred)
```
counter errors_total by app, type

# 5xx requests highlighted as application error
/sortserver '([^\(]+)\(.+ status=(5[0-9]+)/ {
    errors_total["sortserver"][sprintf("%s %s",$1,$2)]++
}
```
 * option B
```
counter errors_total by app, type

# 5xx requests highlighted as application error
/sortserver '([^\(]+)\(.+ status=(5[0-9]+)/ {
    errors_total["sortserver"][$1+" "+$2]++
}
```

I'm basically proposing to add another builtin function `sprintf(...)` wired to golang's `fmt.Sprintf(...)` or at least allow string concatenation with `+` operators.

Thanks in advance for your reply.

lksv:
Adds `concat` function.

It is useful in case when you need construct metric dimension from two parts:

```
counter errors_total by msg

/method=(?P<method>[^ ]+ )status=(?P<status_code>5[0-9]+) / {
    errors_total[concat($1, $2)]++
}
```

fixes #102


lksv:
I have already signed the CLA.

lksv:
I will re-push the the branch in hope, that googlebot will retest the CLA.

freznicek:
Awesome change! Let's get it to master as rc5 ASAP.

jaqx0r:
I think we should do this with the addition operator which is already misused in the parser for pattern concatenation.  Fixing pattern concat to be the same as string concat is required for making #88 and #103 work, such that patterns can appear as part of a conditional expression instead of starting a conditonal by themselves.

Thanks for the patch though.

jaqx0r:
+ now works for string concatenation.

PR #122, opened 2017-11-18
Add Dockerfile

pcarranza:
This makes it easier to test locally by just running it inside a container and throwing logs at it like `docker run --name mtail -v ${HOME}/tmp:/tmp -p 3903:3903 --rm google/mtail -progs /tmp --logs /tmp/<logfile-you-expect>`


pcarranza:
I signed it!

jaqx0r:
Thanks!

PR #123, opened 2017-11-18
Add example file for parsing rails production log

pcarranza:
Parses the total number of started requests by verb, and then completed
requests by status code. It also adds a request latency histogram using
buckets


jaqx0r:
Thanks!

PR #124, opened 2017-11-20
Fix rails sample in regards to parsing errors 500

pcarranza:
An error 500 will print "Internal Server Error" as the description of
the status code, this does not matches `\w+` as the parsing will stop in
the first space character making the whole line to be excluded.

This commit fixes this by using the dot char to match instead.


jaqx0r:
If you want to submit a fragment of a log and expected metrics to the testdata directory, you can add this to the ex_test.go as well.

Issue #127, opened 2017-12-06
Updating a parser file duplicates metrics


ioggstream:
Hi @all, I get the following behavior. Which is the right way to solve it?

### reproduce

1- run mtail  1.0.0-rc4
2- edit a parser file in progs directory

### I expect

1- stats name won't be duplicated

### instead

  * multiple metrics with the same name are shown.
  * prometheus takes only the first one (the older).

```
$ curl 0:3209/metrics |& grep local_total

# TYPE postfix_local_total counter
# postfix_local_total defined at postfix.mtail:20:9-27
postfix_local_total{prog="postfix.mtail"} 2
# postfix_local_total defined at postfix.mtail:20:9-27
postfix_local_total{prog="postfix.mtail"} 0
# postfix_local_total defined at postfix.mtail:20:9-27
portfix_local_total{prog="postfix.mtail"} 104
```

jaqx0r:
Uhoh. Thanks for the report. The oldest metric should be reused but looks
like they are now recreated when the new compile occurs.

ioggstream:
@jaqx0r can you pls point me to the code which is doing the reuse? I could try to fix it...

ioggstream:
The part re-adding metrics is here:

https://github.com/google/mtail/blob/master/metrics/store.go#L35

Is it enough to check that metric labels are not duplicated?

A simple workaround could be to just to prepend new metrics instead of append them.

jaqx0r:
I think the better place to do this is in the loader,
https://github.com/google/mtail/blob/master/vm/loader.go#L198 where we do
the metric add immediately after compilation.  Here instead we should be
testing for existence in the store.

However the complication here is that the loader is only given a list of
metrics to add, and the copmiler has already created new metrics, so the
fix would need to update the program's reference as well.

jprukner:
@ioggstream Any progress on this issue?

ioggstream:
@jprukner no, sorry :(  The submitted patch is working fine months and I hadn't the time to rework on the item, moreover a better solution is quite more complex to implement than mine.

I'm currently working 24/7 on another project: if you can work on that let me know and I'll try to help in the few time I have.

Thanks for asking anyway! Peace, R.

jprukner:
@ioggstream Thanks for response.
[...]

PR #131, opened 2017-12-14
WIP: #127 merge duplicate metrics instead of adding them.

ioggstream:
## what does this PR do

When a new metric is added via store.Add:

  - checks if there's another metric with the same labels
  - if so, copies LabelValues in the new metric
  - pop the old one from the array

### DIFF
--- a/metrics/store.go
+++ b/metrics/store.go
@@ -5,8 +5,10 @@ package metrics
 
 import (
 	"encoding/json"
+	"reflect"
 	"sync"
 
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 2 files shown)


ioggstream:
I signed it!

metrics/store.go
```diff
@@ -26,13 +28,43 @@ func NewStore() (s *Store) {
...
+		// - discard old metric.
+		for i, v := range s.Metrics[m.Name] {
+			//
+			if v.Program != m.Program {
+				continue
+			}
+			if v.Type != m.Type {
+				continue
+			}
+			if reflect.DeepEqual(v.Keys, m.Keys) {
```
ioggstream: @jaqx0r I'm fixing this part up, but I'd appreciate your opinion on that. I spinned it on my test environment and it seems to do the game.

jaqx0r: This seems fine. :)  Thanks!

jaqx0r:
I'd like a test in mtail/mtail_test.go that reproduces the original problem -- creates a program, loads it, checks the exported variables, then does that again, and checks that there are no duplicates in the output.

PR #133, opened 2017-12-18
Create apache_common.mtail

Keleir:
A mtail config for apache common log which extrat the field of %D & %b

The example metric as following:
```
apache_http_response_time{remote_host="127.0.0.1",request_method="GET",request_uri="/server-status?auto",status_code="200",user_agent="Go-http-client/1.1",prog="apache_common.mtail"} 545
```


Keleir:
I signed it!

jaqx0r:
I will assume that this is different to the existing apache compined and apache metrics programs?

PR #138, opened 2018-01-19
set default time zone to "Local"

liyuan35023:
https://github.com/golang/go/blob/7340d1397701a0dc4d2570dac6414f0cdec3fff8/src/time/zoneinfo.go#L280
When Location's name is set to “”，the timezone will be load as UTC.
Therefore, according to the usage of overrideTimezone, I think its default value should be "Local".

### DIFF
--- a/main.go
+++ b/main.go
@@ -67,7 +67,7 @@ var (
 
 	// Runtime behaviour flags
 	syslogUseCurrentYear = flag.Bool("syslog_use_current_year", true, "Patch yearless timestamps with the present year.")
-	overrideTimezone     = flag.String("override_timezone", "", "If set, use the provided timezone in timestamp conversion, instead of the local zone.")
+	overrideTimezone     = flag.String("override_timezone", "Local", "If set, use the provided timezone in timestamp conversion, instead of the local zone.")
 	emitProgLabel        = flag.Bool("emit_prog_label", true, "Emit the 'prog' label in variable exports.")
 
 	// Debugging flags


liyuan35023:
I signed it!

jaqx0r:
Hi, Sorry for the long delay on the review.  Your change makes me realise I don't understrand what is supposed to happen here, emabarrasingly.

I'm not convinced your change is correct;  perhaps the comment should be changed to say it's UTC unless you override it.

I don't want a surprise change on existing users who assume it's UTC because they didn't read the comment and things just worked until now.  I think the correct thing to do here is to update the flag description and put a note in the docs (on Deployment?) about the flag and suggested values.

main.go
```diff
@@ -67,7 +67,7 @@ var (
 
 	// Runtime behaviour flags
 	syslogUseCurrentYear = flag.Bool("syslog_use_current_year", true, "Patch yearless timestamps with the present year.")
-	overrideTimezone     = flag.String("override_timezone", "", "If set, use the provided timezone in timestamp conversion, instead of the local zone.")
+	overrideTimezone     = flag.String("override_timezone", "Local", "If set, use the provided timezone in timestamp conversion, instead of the local zone.")
```
jaqx0r: As noted, please leave the value but update the description, and optionally add a comment to the docs.  (I can do this if you aren't sure where to.)

jaqx0r:
I've submitted 2ff438e instead which clarifies the flag and adds some documentation.

PR #141, opened 2018-01-24
Fix tl;dr building recipe and some links

darkk:
Otherwise `go get` from go1.9.2 failed with

```
$ go get https://github.com/google/mtail
package https:/github.com/google/mtail: "https://" not allowed in import path
```


jaqx0r:
Thanks!

PR #144, opened 2018-02-27
Adding cross-building to Makefile & Travis

jamtur01:
1. Adds crossbuilds for amd64 on Linux/Darwin/Windows using gox
2. Adds deploy step to both Travis and Circle since I couldn't work out which is current. The deploy step only triggers on commits with tags.
3. You'll need to specify GH api keys for both to make this work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,7 @@ covclean:
 
 version := $(shell git describe --tags)
 revision := $(shell git rev-parse HEAD)
+release := $(shell git describe --tags | cut -d"-" -f 1,2)
 
 install mtail: $(GOFILES) .dep-stamp
 	go install -ldflags "-X main.Version=${version} -X main.Revision=${revision}"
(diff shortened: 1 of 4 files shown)


jamtur01:
The Circle failure seems unrelated to this? Something about coveralls?

jaqx0r:
Thanks for the PR!  You're the winner of First PR Since We Set Up Circle so I'm not surprised it's unexpected.  The intent was to warn if test coverage goes down.

jaqx0r:
The Coveralls failure looks like a side effect of this being a PR, not a postcommit run.  We can ignore it for now.

Makefile
```diff
@@ -40,6 +41,13 @@ vm/parser.go: vm/parser.y .gen-dep-stamp
 emgen/emgen: emgen/emgen.go
 	cd emgen && go build
 
+install_gox:
+	go get github.com/mitchellh/gox
+
+crossbuild: install_gox $(GOFILES) .dep-stamp
```
jaqx0r: Insert surprised face emoji here.  Wow.  I didn't realise it was this easy!  Thanks.

jamtur01:
I didn't test the deploy step because of auth reason s. So that's an exiting surprise for next release. 😁

PR #149, opened 2018-04-03
Fix linter issues

mjtrangoni:
Hi @jaqx0r ,

I fixed some trivial issues reported by *megacheck*, *maligned*, *golint*, and *vetshadow*. I signed the CLA already. 

Thanks!

I will be double-checking the CI-tests tomorrow, and fix them.

### DIFF
--- a/mtail/mtail.go
+++ b/mtail/mtail.go
@@ -144,23 +144,22 @@ func (m *MtailServer) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // Options contains all the parameters necessary for constructing a new MtailServer.
 type Options struct {
+	BindAddress          string
 	Progs                string
+	BuildInfo            string
 	LogPathPatterns      []string
 	LogFds               []int
-	BindAddress          string
 	OneShot              bool
 	CompileOnly          bool
 	DumpAst              bool
 	DumpAstTypes         bool
 	DumpBytecode         bool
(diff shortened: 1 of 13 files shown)


jaqx0r:
Thanks!  I'm not certain about the rename of MtailServer to Server and the removal of the two flags, but the rest looks fine.  Let me know once the CI passes.

mjtrangoni:
@jaqx0r 
I drop the MtailServer to Server renaming. About the 2 flags, they are not being used at all anywhere, and `megacheck` complains about it. By the way, the 12 CI checks are happy now.  

It would be nice to have this fixed too, see,
```
$ gometalinter --vendor --disable-all --enable=golint ./... | grep exported | wc -l
53
```
Everything exported should have a comment or be unexported.

jaqx0r:
Thanks for fixing the CI tests.

I'd like to remove the flags independently of a megacheck fix, as they're
remaining only because I don't have any good way to deprecate the flag
usage safely.  If you please, can you remove the removal here and we can do
it in a separate change.

I've been working on and off documenting every exported symbol, but only as
I am revisiting code and noticing.  I have gometalinter running on all code
I open in emacs at the moment and try to fix things as I come across them.

mjtrangoni:
@jaqx0r both flags are back there.

jaqx0r:
Thanks for indulging me.

PR #152, opened 2018-04-09
[docs] change time.Parse example default format

hkwi:
03 format is for Kitchen

```
package main

import (
    "time"
    "fmt"
)

func main(){
    fmt.Println(time.Parse("2006/01/02 03", "2018/04/09 00"))
    // 2018-04-09 00:00:00 +0000 UTC <nil>
    fmt.Println(time.Parse("2006/01/02 03", "2018/04/09 12"))
    // 2018-04-09 12:00:00 +0000 UTC <nil>
    fmt.Println(time.Parse("2006/01/02 03", "2018/04/09 13"))
    // 0001-01-01 00:00:00 +0000 UTC parsing time "2018/04/09 13": hour out of range
}
```

### DIFF
--- a/docs/Programming-Guide.md
+++ b/docs/Programming-Guide.md
@@ -77,7 +77,7 @@ def syslog {
         }
         # If the RFC3339 style matched, parse it this way.
         len($rfc3339_date) > 0 {
-            strptime($rfc3339_date, "2006-01-02T03:04:05-0700")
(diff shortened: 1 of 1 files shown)


jaqx0r:
Wow.  Thanks!  I wonder how many of my other tests are wrong now.

PR #156, opened 2018-04-19
Docker Hub tagging for Semantic Versioning

jnovack:
This PR completes the Docker Hub building (if/when it happens).

Upon a successful push to `master`, Docker Hub will build and tag all versions appropriately as long as you continue to tag as you currently do.  This makes sure that older versions are always kept, and tags are updated appropriately.

An example is at [jnovack/docker-hub-test](https://hub.docker.com/r/jnovack/docker-hub-test/tags/).  As you can see:

* When **v1.0.0-alpha2** is pushed, the following tags are updated, `1`, `1.0`, `1.0.0`, `latest`.
* When **v1.0.0-rc1** is pushed, the following tags are updated, `1`, `1.0`, `1.0.0`, `latest`. `1.0.0-alpha2` remains untouched.
* When **v1.0.0** is pushed, the following tags are updated, `1`, `1.0`, `1.0.0`, `latest`. `1.0.0-alpha2` and `1.0.0-rc1` remains untouched.

Expanding...

* When **v1.0.1** is pushed,  the following tags are updated, `1`, `1.0`, `1.0.1`, `latest`.  `1.0.0` remains untouched.
* When **v1.1.0** is pushed,  the following tags are updated, `1`, `1.1`, `1.1.0`, `latest`.  `1.0.0`, `1.0.1` and `1.0` remain untouched.

Version 1.0.0-alpha2 is pushed.
image

Version 1.0.0-rc1 is pushed.
image

Version 1.0.0 is pushed.
image

It is HEAVILY recommended to push the tag BEFORE pushing the commit (`git push --tags` THEN `git push`).   Yes, Docker does wait a few minutes before building, but you don't want Docker to build before it has the right tag.

I follow [semver](https://semver.org) guidelines when searching for tags.  To ensure that the `post_build` script finds them, you [MUST](http://tools.ietf.org/html/rfc2119) use a lowercase "v" followed by three (3) sets of digits taking the form of `X.Y.Z`.  Pre-release builds suffix the version with a hyphen ("-") and any amount of alpha-numeric ("0-9, A-Z, a-z").


jaqx0r:
Thanks!

PR #158, opened 2018-05-21
Minor typo in example "five nines" vs "nine fives"

rapaul:
The example walks through an example where the request time SLI is 0.555555555 (nine fives), but the last sentence confuses the SLI and wording reverts to five nines.


jaqx0r:
It was supposed to be a bad joke.  I understand that it's confusing instead
:)

rapaul:
This PR just makes sure the bad joke (nine fives) had continuity right to the last paragraph :D
Thanks for the quick merge.

PR #163, opened 2018-06-26
Correctly handle duplicated metrics

pborzenkov:
The PR addresses two problems with duplicated metrics:
  - counters w/o labels weren't correctly handled, as even an empty new counter already contains a single value (zero);
  - metrics with labels weren't handled by dedup at all.


pborzenkov:
I signed it!

jaqx0r:
thanks!

PR #164, opened 2018-06-27
vm: load programs right after watcher.Create event

pborzenkov:
watcher.Update event might not be received when a new file is added into
progs directory. For example, when RPM is installing a package, it
doesn't change mtime on the files, thus watcher.Update event is never
received on the actual program file and the programs installed by RPM
are never loaded.

I0627 12:11:59.401242  840953 log_watcher.go:89] watcher event "/etc/mtail/messages.mtail;5b3354df": RENAME
I0627 12:11:59.401267  840953 log_watcher.go:89] watcher event "/etc/mtail/messages.mtail": CREATE

Fix this by loading programs right after watcher.Create event.


jaqx0r:
Thanks!

PR #175, opened 2018-09-20
vm timeMemos on LRU cache

hkwi:
Process memory grows because timeMemos in VM holds all reference to unique time spec strings in the log. Using LRU cache saves memory consumption.


jaqx0r:
Great, thanks!

Issue #176, opened 2018-09-20
glob not working for files created POST startup


jnovack:
Globbing does not appear to be watching files that are created AFTER **mtail** startup.

```
# test.mtail
counter line_count
/$/ {
    line_count++
}
```

```
$ mtail -logs log.*.txt --vmodule=log_watcher=2 -progs . -logtostderr -poll_interval 1s
I0920 11:17:29.154842   52436 main.go:93] mtail version v3.0.0-rc16 git revision 77eeba903db0ffd90e29c1ee778fdefb0112bfab go version go1.11
I0920 11:17:29.154934   52436 main.go:94] Commandline: ["mtail" "-logs" "log.*.txt" "--vmodule=log_watcher=2" "-progs" "." "-logtostderr" "-poll_interval" "1s"]
I0920 11:17:29.155105   52436 log_watcher.go:162] Resolved path for lookup "/Users/jnovack/working/mtail-test"
I0920 11:17:29.155140   52436 log_watcher.go:137] Adding a watch on resolved path "/Users/jnovack/working/mtail-test"
I0920 11:17:29.155593   52436 mcp.go:220] Loaded program test.mtail
I0920 11:17:29.155605   52436 mcp.go:239] Program test.mtail has goroutine marker 0x74657374
I0920 11:17:29.155644   52436 vm.go:707] Starting program test.mtail
I0920 11:17:29.155656   52436 mcp.go:243] Started test.mtail
I0920 11:17:29.155768   52436 log_watcher.go:162] Resolved path for lookup "/Users/jnovack/working/mtail-test"
I0920 11:17:29.155875   52436 mtail.go:338] Listening on port :3903
## touch log.1.txt
I0920 11:17:37.417408   52436 log_watcher.go:89] watcher event "/Users/jnovack/working/mtail-test/log.1.txt": CREATE
I0920 11:17:37.417504   52436 log_watcher.go:162] Resolved path for lookup "/Users/jnovack/working/mtail-test/log.1.txt"
I0920 11:17:37.417514   52436 log_watcher.go:137] Adding a watch on resolved path "/Users/jnovack/working/mtail-test/log.1.txt"
## cat test.mtail >> log.1.txt
I0920 11:17:50.635282   52436 log_watcher.go:89] watcher event "/Users/jnovack/working/mtail-test/log.1.txt": WRITE
I0920 11:17:52.419810   52436 log_watcher.go:89] watcher event "/Users/jnovack/working/mtail-test/log.1.txt": CHMOD
[...]

PR #177, opened 2018-09-20
Update log_glob_test.sh

jnovack:
The log_glob_test.sh test was incorrect causing a false positive pass for glob testing.  Additionally, the false positive had a side-effect that files created AFTER startup were not being properly monitored.

This only fixes the test, this does NOT fix the main issue as discussed in #176.

On line 19, when you create `log1`, the total number of logs should be "2", `log` and `log1`.  Additionally, this also tests that the line_counts are properly incremented, especially when a file that is being globbed was created post startup.

This fails because well, `tailer/tail.go` is incorrect and right now the fix is beyond my comprehension.

### DIFF
--- a/tests/log_glob_test.sh
+++ b/tests/log_glob_test.sh
@@ -15,11 +15,13 @@ sleep 1
 
 uri_get /debug/vars
 expect_json_field_eq "1" log_count "${DATA}"
+expect_json_field_eq "1" line_count "${DATA}"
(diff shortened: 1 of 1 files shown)


jaqx0r:
facepalm emoji.

Thanks fro the fixed test. I'm going to merge this because I'd rather have a broken CI than release broken code.

PR #186, opened 2018-10-31
suppress redundant source comments

zarath:
Only print out source comments in prometheus exporter if they differ from previous one.


zarath:
I signed it!

zarath:
Non intended debian files ....

Issue #202, opened 2019-01-16
del after duration / expire not working 


chaudhryfaisal:
### mtail version
```shell
mtail version  git revision  go version go1.11.4 go arch amd64 go os darwin
git revision f8eca1029c15a6a9d3c125bb9d5771a0a6cc4b42 
```

#### what happened
trying to test expire metric by following **linecount.mtail** :

```
counter line_count_1 by a
counter line_count_2

/$/ {
  line_count_1["a"]++
  del line_count_1["a"] after 30s

  line_count_2++
  del line_count_2 after 30s
}

```

```shell
I0115 21:57:33.034933   56251 main.go:91] mtail version  git revision  go version go1.11.4 go arch amd64 go os darwin
I0115 21:57:33.035145   56251 main.go:92] Commandline: ["/private/var/folders/sd/n7x4tn6j0vz9jqn00w901ss95pmzml/T/___run" "--logtostderr" "--progs" "mtail/__test/linecount.mtail" "--logs" "mtail/__test/sample.log"]
I0115 21:57:33.035949   56251 loader.go:219] Loaded program linecount.mtail
I0115 21:57:33.035959   56251 loader.go:238] Program linecount.mtail has goroutine marker 0x6c696e65
I0115 21:57:33.035992   56251 vm.go:749] Starting program linecount.mtail
I0115 21:57:33.035998   56251 loader.go:242] Started linecount.mtail
I0115 21:57:33.036428   56251 tail.go:254] Tailing mtail/__test/sample.log
I0115 21:57:33.036544   56251 mtail.go:251] Listening on [::]:3903
```

#### browser output
```
# TYPE line_count_1 counter
# line_count_1 defined at linecount.mtail:1:9-20
line_count_1{a="a",prog="linecount.mtail"} 3
# TYPE line_count_2 counter
# line_count_2 defined at linecount.mtail:2:9-20
line_count_2{prog="linecount.mtail"} 3
```

#### expected behavior
line_count_1 and line_count_2 to be deleted after `30s`

#### actual behavior
none of the metrics expire

jaqx0r:
I noticed this a few days ago, it's fixed in db51966fd590ba1f4904b57859adaca6a7c05d27 and in the just released https://github.com/google/mtail/releases/tag/v3.0.0-rc18

Please get the latest and let me know if you still have this problem, thanks!

chaudhryfaisal:
[...]

PR #205, opened 2019-01-30
adding option to configure StoreExpireTickInterval and Fixed bug where `Store.Expire()` was only invoked once

chaudhryfaisal:
adding option to configure StoreExpireTickInterval
Fixed bug where `Store.Expire()` was only invoked once

### DIFF
--- a/internal/metrics/store.go
+++ b/internal/metrics/store.go
@@ -102,6 +102,7 @@ func (s *Store) MarshalJSON() (b []byte, err error) {
 // Expire iterates through the Store looking for metrics that have been marked
 // for expiry, and removing them if their expiration time has passed.
 func (s *Store) Expire() error {
+	glog.Info("Running Store.Expire()")
 	s.Lock()
 	defer s.Unlock()
 	now := time.Now()
@@ -124,14 +125,17 @@ func (s *Store) Expire() error {
 }
 
 // StartExpiryLoop runs a permanent goroutine to expire metrics every hour.
-func (s *Store) StartExpiryLoop() {
+func (s *Store) StartExpiryLoop(duration time.Duration) {
 	go func() {
-		ticker := time.NewTicker(time.Hour)
-		select {
-		case <-ticker.C:
-			err := s.Expire()
-			if err != nil {
-				glog.Info(err)
+		glog.Infof("Configuring StartExpiryLoop with %s", duration.String())
+		ticker := time.NewTicker(duration)
+		for {
+			select {
+			case <-ticker.C:
+				err := s.Expire()
+				if err != nil {
+					glog.Info(err)
+				}
 			}
 		}
 	}()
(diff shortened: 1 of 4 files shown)


jaqx0r:
Thanks. Can you resolve the clabot comments?

chaudhryfaisal:
> Thanks. Can you resolve the clabot comments?

Done

internal/mtail/mtail.go
```diff
@@ -344,7 +345,7 @@ func (m *Server) Run() error {
 			return err
 		}
 	} else {
-		m.store.StartExpiryLoop()
+		m.store.StartExpiryLoop(*m.storeExpireTickInterval)
```
jaqx0r: The test failure from CI is because this is nil when run in test.
```suggestion
                if (*m.storeExpireTickInterval != nil ) {
		  m.store.StartExpiryLoop(*m.storeExpireTickInterval)
		}
```

internal/mtail/mtail.go
```diff
@@ -58,9 +58,10 @@ type Server struct {
 	dumpAstTypes bool // if set, mtail prints the program syntax tree after type checking
 	dumpBytecode bool // if set, mtail prints the program bytecode after code generation
 
-	syslogUseCurrentYear bool // if set, use the current year for timestamps that have no year information
-	omitMetricSource     bool // if set, do not link the source program to a metric
-	omitProgLabel        bool // if set, do not put the program name in the metric labels
+	syslogUseCurrentYear    bool           // if set, use the current year for timestamps that have no year information
+	omitMetricSource        bool           // if set, do not link the source program to a metric
+	omitProgLabel           bool           // if set, do not put the program name in the metric labels
+	storeExpireTickInterval *time.Duration // interval to expire/delete metrics from store
```
jaqx0r: You could not have a pointer here and instead leave it as the zero value, and then we can change my suggested `if != nil` to if duration > 0 and put it inside StartExpiryLoop() instead.

internal/metrics/store.go
```diff
@@ -124,14 +125,17 @@ func (s *Store) Expire() error {
...
-func (s *Store) StartExpiryLoop() {
+func (s *Store) StartExpiryLoop(duration time.Duration) {
 	go func() {
-		ticker := time.NewTicker(time.Hour)
-		select {
-		case <-ticker.C:
-			err := s.Expire()
-			if err != nil {
-				glog.Info(err)
+		glog.Infof("Configuring StartExpiryLoop with %s", duration.String())
```
jaqx0r: e.g. 

```suggestion
if duration <= 0 {
glog.Infof("Metric store expiration disabled")
return
}
```

internal/mtail/mtail.go
jaqx0r: This is nil when the test fails, because internal/mtail/basic_tail_integratoin_test.go and others do not initialize the expiration option.  They all could (in internal/mtail/testing.go:MakeTestServer but I think it's better to just handle it, see above.)

chaudhryfaisal: Thanks I couldn't really figure out where the nill was coming from. I tried to look at the logs or run test locally but didn't have much luck.

jaqx0r:
Fixes #202

PR #215, opened 2019-03-07
Add PREFIX to make install

worxli:
Commit #3bfdc27 broke the Dockerfile. I.e. `COPY --from=builder /go/bin/mtail /usr/bin/mtail` is trying to copy from a non-existent file.


worxli:
I signed it!

jaqx0r:
Thanks! Docker builds don't get tested by ci yet

PR #218, opened 2019-03-14
Microsecond to second conversion was wrong.

jakirpatel:
Please check the conversion. Ideally it should be ,
 $time_us / 1000000 or $time_us * 0.000001


jakirpatel:
I signed it!

jaqx0r:
Doh! Thanks

PR #231, opened 2019-05-14
histogram example improvement

Alb0t:
Looking at the histogram documentation and examples, I felt a bit lost at first. In hindsight, it all seems so simple. Wanted to add some depth to the example so nobody else stumbles on the silly stuff like me.


Alb0t:
I signed it!

jaqx0r:
Awesome, thankyou!

PR #236, opened 2019-06-12
Feature/gomod

ayjayt:
This commit adds a go.mod, but turns off GO111MODULE globally (especially for
dependency building, which would expect full paths with mod enabled). It turns
GO111MODULE back on for the actual target build stage, allowing go to
use import paths relative to the working directory (not GOPATH),
especially internal paths. Although this isn't a full migration to go mod, it does allow us to build outside of GOPATH.

NOTE: I did not test _all_ the Makefile commands for compatibility with go.mod, so the Makefile disables go.mod except in the explicit instances where it's used. If GO111MODULE is set explicitly, the Makefile will not turn it off.


ayjayt:
Just tagging @jaqx0r in case you don't get notified!

jaqx0r:
Super cool, thanks!

PR #241, opened 2019-07-05
[examples] Fixed postfix matching with submission

hp197:
* Added forward slash matching in application pattern
  * Made a regex of smtpd matching to allow for "submission/smtpd"


hp197:
I signed it!

jaqx0r:
Thanks!

PR #244, opened 2019-07-23
Use consistent naming for internal metrics

SuperQ:
Make the internal metric names consistent, follow Prometheus/OpenMetrics
naming conventions.


SuperQ:
There's still `line_count` which I would like to rename to `lines_total`, but the use is sprinkled all over the codebase.

jaqx0r:
Great, thanks!

PR #245, opened 2019-07-23
Add missing filter to container workflow

SuperQ:
The `container` phase of the workflow was missing tags so it was never
run, which causes the release to not be published.


SuperQ:
It looks like without the tag filter, the `container` workflow was only running on master, not tags.

jaqx0r:
:man_facepalming:

PR #258, opened 2019-08-26
doc: fix date format layout

antong:
Fix the RFC3339 time parsing layout. The regexp correctly matched the time zone offset (07:00), but the time parsing layout incorrectly left out the colon (0700).


jaqx0r:
Thanks!

PR #266, opened 2019-08-31
make timestamp() return float64

### LINKED ISSUE #263: Timestamp resolution
Mtail records timestamps as Go time.Time values, that have nanosecond resolution. However, when accessing a time acquired using `timestamp()` from a program, e.g., when subtracting two timestamps or assigning a timestamp to a gauge, then the timestamps are converted to integers using `Time.Unix()`. This means that the resolution when comparing timestamps is 1s. This can be ok when monitoring longer sessions such as in the rsync example, where the durations are tens of seconds or more. But if the durations are in single digit seconds, the systematic error is significant. Many applications have sub-second transactions or sessions, and these will all be rounded to 0.

Here is a toy demonstration that monitors the duration between emitted log lines for the mtail log itself.
```
# demo.mtail
counter lines_total

# For why the 0 and negative buckets are needed, see:
# https://github.com/google/mtail/issues/226#issuecomment-524862192
histogram delta_t buckets -1e9, 0, 0.0005, 0.001, 0.01, 0.1, 1, 10, 100

gauge last

def glog {
        /(?P<level>[IWECF])(?P<timestamp>\d{4} \d{2}:\d{2}:\d{2}\.\d{6}) +/
        /(?P<thread>\d+) (?P<src_file>[^:]+):(?P<src_line>\d+)\] / {
                strptime($timestamp, "0102 15:04:05")
                next
        }
}

@glog {

/$/ {
        lines_total > 0 {
                delta_t = (timestamp() - last)
        }
        last = timestamp()
        lines_total++
}

}
``` 

Example input:
```
I0829 22:08:41.220692    1288 main.go:100] mtail version invalid ... go version go1.12.7 go arch amd64 go os windows
I0829 22:08:41.221691    1288 main.go:101] Commandline: ["C:\\mtail.exe" "-progs" "demo.mtail" "-logs" "foo.log" "-logtostderr"]
I0829 22:08:41.224689    1288 log_watcher.go:248] No abspath in watched list, added new one for X
I0829 22:08:41.224689    1288 checker.go:202] declaration of capture group reference `src_file' at demo.mtail:7:2-59 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `src_line' at demo.mtail:7:2-59 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `src_line' at demo.mtail:7:2-59 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `thread' at demo.mtail:7:2-59 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `thread' at demo.mtail:7:2-59 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `src_file' at demo.mtail:7:2-59 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `level' at demo.mtail:6:2-68 appears to be unused
I0829 22:08:41.225714    1288 checker.go:202] declaration of capture group reference `level' at demo.mtail:6:2-68 appears to be unused
I0829 22:08:41.226713    1288 loader.go:229] Loaded program demo.mtail
I0829 22:08:41.226713    1288 log_watcher.go:248] No abspath in watched list, added new one for X
W0829 22:08:41.227713    1288 mtail.go:101] No matches for pattern "foo.log"
I0829 22:08:41.227713    1288 store.go:136] Starting metric store expiry loop every 1h0m0s
I0829 22:08:41.227713    1288 tail.go:435] Starting log handle expiry loop every 1h0m0s
I0829 22:08:41.227713    1288 mtail.go:321] Listening on [::]:3903
I0829 22:08:48.037569    1288 log_watcher.go:253] Found this processor in watched list
I0829 22:08:48.038569    1288 log_watcher.go:248] No abspath in watched list, added new one for C:\foo.log
I0829 22:08:48.044581    1288 tail.go:289] Tailing C:\foo.log
I0829 22:09:21.933622    1288 mtail.go:349] Received SIGTERM, exiting...
I0829 22:09:21.934635    1288 mtail.go:363] Shutdown requested.
I0829 22:09:21.935667    1288 log_watcher.go:218] Shutting down log watcher.
I0829 22:09:21.935667    1288 loader.go:381] Shutting down loader.
I0829 22:09:21.936655    1288 mtail.go:386] END OF LINE
```

Example Prometheus metrics:
```
delta_t_bucket{prog="demo.mtail",le="0"} 22
delta_t_bucket{prog="demo.mtail",le="0.0005"} 22
delta_t_bucket{prog="demo.mtail",le="0.001"} 22
delta_t_bucket{prog="demo.mtail",le="0.01"} 22
delta_t_bucket{prog="demo.mtail",le="0.1"} 22
delta_t_bucket{prog="demo.mtail",le="1"} 22
delta_t_bucket{prog="demo.mtail",le="10"} 23
delta_t_bucket{prog="demo.mtail",le="100"} 24
delta_t_bucket{prog="demo.mtail",le="+Inf"} 24
delta_t_sum{prog="demo.mtail"} 40
delta_t_count{prog="demo.mtail"} 24
```

Here we see that all durations were counted as 0, except for the 7s and 33s pauses, that add up to the sum of 40s.

I tested and changed the few places in vm.go that use Unix seconds to use Unix nanoseconds instead. This of course works, but is hardly backwards compatible as `timestamp()` returns a billion times larger values. Perhaps changing to Unix seconds as a float64 would be better, as that wouldn't change the unit and would give microsecond resolution.

antong:
This changes the return type of the built-in timestamp() function to
float64 instead of int64. The unit remains the same, that is seconds
since the Unix epoch. Before this change, timestamps were truncated to
integer seconds, which made it impossible to measure durations less than
a second. Unix seconds as a float64 give better than microsecond
resolution.

Fixes #263

### DIFF
--- a/internal/mtail/testdata/rsyncd.golden
+++ b/internal/mtail/testdata/rsyncd.golden
@@ -1,6 +1,6 @@
 counter bytes_total {operation=sent} 62793673 2011-02-23T05:54:10Z
 counter bytes_total {operation=received} 975017 2011-02-23T05:54:10Z
 counter connections_total 52 2011-02-22T21:54:13Z
-counter connection-time_total 1181011 2011-02-23T05:54:10Z
+counter connection-time_total 1181011.0 2011-02-23T05:54:10Z
 counter transfers_total {operation=send,module=module} 2 2011-02-23T05:50:32Z
 counter transfers_total {operation=send,module=repo} 25 2011-02-23T05:51:14Z
--- a/internal/vm/types/types.go
+++ b/internal/vm/types/types.go
@@ -202,7 +202,7 @@ var Builtins = map[string]Type{
 	"bool":        Function(NewVariable(), Bool),
 	"float":       Function(NewVariable(), Float),
 	"string":      Function(NewVariable(), String),
-	"timestamp":   Function(Int),
+	"timestamp":   Function(Float),
 	"len":         Function(String, Int),
 	"settime":     Function(Int, None),
 	"strptime":    Function(String, String, None),
--- a/internal/vm/vm.go
+++ b/internal/vm/vm.go
@@ -116,6 +116,18 @@ func (v *VM) errorf(format string, args ...interface{}) {
 	v.terminate = true
 }
 
+// unixFloat returns t as a Unix time, the number of seconds elapsed
+// since January 1, 1970 UTC.
+func unixFloat(t time.Time) float64 {
+	return float64(t.Unix()) + float64(t.Nanosecond())/1e9
+}
+
+// timeFloat returns the time.Time corresponding to the given Unix time.
+func timeFloat(unix float64) time.Time {
+	sec, frac := math.Modf(unix)
+	return time.Unix(int64(sec), int64(frac*1e9))
+}
+
 func (t *thread) PopInt() (int64, error) {
 	val := t.Pop()
 	switch n := val.(type) {
@@ -143,6 +155,8 @@ func (t *thread) PopFloat() (float64, error) {
 	case float64:
 		return n, nil
 	case int:
+		return float64(n), nil
+	case int64:
 		return float64(n), nil
 	case string:
 		r, err := strconv.ParseFloat(n, 64)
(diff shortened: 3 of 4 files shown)


antong:
I couldn't figure out how to declare in the golden format that a metric that has no value is a float.

jaqx0r:
I appreciate you taking the time to make the PR but I think changing time from int to float isn't the right fix; I think it something that looks like time.Time should be exposed in the mtail language.

antong:
Yes, I understand. This PR was more to just open the discussion and experiment what the minimal change that timestamp() returned a float would look like, and to see how backwards compatible it is. And it actually quite does work. Existing programs automatically get improved resolution, and even the tests suite mostly passes (couldn't figure out how to articulate that a metric is a float in the test result "golden" format.)

It clearly would be more elegant if there was a Time type, but I don't see how that could be backwards compatible with existing mtail programs in case that is a requirement. Perhaps by making a timestamp2() function that returns the new type? If there is a Time type, it might also benefit from a Duration/Interval type as there is in Go, if we want to distinguish between calendar time and intervals using type safety.

Or, perhaps nanotimestamp() function, that is like timestamp() but returns nanoseconds since epoch instead of seconds, could be an acceptable stopgap for new programs that require higher resolution?

jaqx0r:
For public (and my own) reference, I'm leaving this PR open to remind me to fix this.

jaqx0r:
Changed my mind, the bug that's open is sufficient to remind me to do this.

PR #277, opened 2019-11-21
Improve documentation of timestamp handling

SuperQ:
Change the wording of the timestamp handling documentation. The
current wording is incorrect and is misleading and confusing users.


jaqx0r:
Thanks!

PR #288, opened 2020-01-20
Update build.

SuperQ:
* Update to Go 1.13.
* Update all Go dependencies.

Signed-off-by: Ben Kochie <superq@gmail.com>


SuperQ:
I'm not sure what's up with the coverage build.

jaqx0r:
The coverage failure is why I haven't upgraded to 1.13 yet.

SuperQ:
https://github.com/golang/go/issues/35781 :frowning_face: 

This coverage tool seems to break often.  Maybe try https://codecov.io/ instead? The free plan should be enough for mtail.

jaqx0r:
The problem is with go cover, not with the coverage hosting, unless I've
misread what you're suggesting?

SuperQ:
Yes, I'm suggesting that the coverage tool is not stable enough for use, and that it's worth considering other tools, hosted or not.

jaqx0r:
Oh I see, with codecov.io we don't need to run go tool cover to process the
coverage reports generated by go test?

In fact, I think we don't need go tool cover in mtail either -- If I'm not
mistaken now I've reread the Makefile, it's only being used to generate an
html report, which we don't really need at all.

jaqx0r:
I've removed the errant command from the coverage step in CircleCI's config now, so this should go through fine.

jaqx0r:
Sigh, fine, this needs more work.  Going to disable coverage for now and file an issue to fix it.

SuperQ:
Thanks for merging. That sucks, good luck with the coverage tool issues.

PR #296, opened 2020-02-12
Set exit code to 0 when outputting version info.

### LINKED ISSUE #295: Do not exit with status code 1 when given -version flag
I am maintaining a tool that helps install and update commandline utilities distributed as single binaries through Github, just like mtail is. (The tool is called [vk](https://github.com/cellpointmobile/vk)) This is done in part by parsing the version output from the utility itself. However, mtail does something that no other utility seems to do, it exits with status code 1, instead of 0 when given the -version flag. As you probably are aware, any exit status code above 0 is considered an error in Posix programs. Therefore vk can not, without ugly hacks, handle mtail.
Would you be open to having mtail exit with status code 0 when outputting version info?

drzero42:
Fixes #295

### DIFF
--- a/cmd/mtail/main.go
+++ b/cmd/mtail/main.go
@@ -95,7 +95,7 @@ func main() {
 	flag.Parse()
 	if *version {
 		fmt.Println(buildInfo.String())
-		os.Exit(1)
+		os.Exit(0)
 	}
 	glog.Info(buildInfo.String())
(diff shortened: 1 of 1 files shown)


drzero42:
@googlebot I signed it!

drzero42:
@googlebot I fixed it.

jaqx0r:
Thanks for the contribution!

Issue #273, opened 2019-10-31
"failed to marshal metrics into json" due to +Inf


gsstark:
I'm getting this error when running mtail with -one_shot mode:

```
Metrics store:E1031 11:27:10.506026   50999 main.go:179] failed to marshal metrics into json: json: unsupported value: +Inf
```

I started getting this error when I first added a histogram which makes me think it has to do with the le=+Inf  label. But I haven't dug any deeper yet.

jaqx0r:
Can you share the whole command line please?
Did you do anything other than run the command?

JSON is not well supported, it started as a hack to test metrics and needs
to be removed. I wanna ask, do you care particularly about the format of
results from one_shot?

Also it'd help make a test case to reproduce if you can share the program
you're writing.

Lusitaniae:
I see the same

nginx.mtail
```
counter http_requests_total by method, code, content_type
counter http_request_duration_milliseconds_sum by method, code, content_type
counter http_response_size_bytes_sum by method, code, content_type
histogram http_request_duration_miliseconds by method, code buckets 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 15

/^/ +
/(?P<proxy_addr>[0-9A-Za-z\.\-:]+) / +
/- / +
/(?P<remote_user>[0-9A-Za-z\-]+) / +
/(?P<time_local>\[\d{2}\/\w{3}\/\d{4}:\d{2}:\d{2}:\d{2} \+\d{4}\]) / +
/"(?P<request_method>[A-Z]+) (?P<request_uri>\S+) (?P<http_version>HTTP\/[0-9\.]+)" / +
/(?P<status>\d{3}) / +
/(?P<bytes_sent>\d+) / +
/(?P<request_seconds>\d+)\.(?P<request_milliseconds>\d+) / +
/"(?P<http_referer>\S+)" / +
/"(?P<http_user_agent>[[:print:]]+)" / +
/"(?P<content_type>.+)" / +
/"(?P<remote_addr>[0-9A-Za-z\.\-:]+)"/ +
/$/ {
  http_requests_total[tolower($request_method)][$status][$content_type]++
  http_request_duration_milliseconds_sum[tolower($request_method)][$status][$content_type] += $request_seconds * 1000 + $request_milliseconds
  http_request_duration_miliseconds[tolower($request_method)][$status] = $request_seconds * 1000 + $request_milliseconds
[...]

PR #304, opened 2020-03-28
MarshalJSON bucket bounds as strings

quatre:
When outputting JSON (with -one_shot for example), the +Inf bucket bound
could not be marshaled as a JSON number and caused mtail to crash.

This stringifies the buckets bounds and solves this issue.
Resolves: https://github.com/google/mtail/issues/273


quatre:
@googlebot I signed it!

jaqx0r:
Thanks!

Issue #305, opened 2020-03-29
Providing a flag to hide log lines from the web interface


quatre:
Hello!

currently, mtail can expose log lines through the web interface (for example through the /progz endpoint) when they cause a program to fail (eg. by trying to compare a string with an int).

What is your opinion on having a command line flag that would prevent these log lines to be exposed through the web interface?

jaqx0r:
Why do you want to turn that feature off?

quatre:
In some cases, logs may contain sensitive information (think for example, a username, or an IP address) that would be better not to expose.

When things are working nicely, this shouldn't be a problem. But if the application log format changes  and breaks the mtail program, it could become an information leak.

jaqx0r:
Yep that makes sense. Do you not firewall the mtail port though?

quatre:
We sure would firewall it :p!
But the people who have shell access to Prometheus (to edit its config for example) do not necessarily have access to sensitive logs.

jaqx0r:
That's a good reason!

robert-heinzmann-logmein:
I reopened the original PR with a new branch. I also added an option to disable the /varz and /progz endpoints, as those could also leak some information that is not intended for public access as typically needed with prometheus.

PR #308, opened 2020-04-03
Add a -debug flag to enable debug endpoints

quatre:
This would implement https://github.com/google/mtail/issues/305 but we can surely improve it as I (lazily, I confess) put many different endpoints under the same flag.

### DIFF
--- a/cmd/mtail/main.go
+++ b/cmd/mtail/main.go
@@ -47,6 +47,7 @@ var (
 	dumpAst      = flag.Bool("dump_ast", false, "Dump AST of programs after parse (to INFO log).")
 	dumpAstTypes = flag.Bool("dump_ast_types", false, "Dump AST of programs with type annotation after typecheck (to INFO log).")
 	dumpBytecode = flag.Bool("dump_bytecode", false, "Dump bytecode of programs (to INFO log).")
+	debug        = flag.Bool("debug", false, "Enable debugging endpoints.")
 
 	// VM Runtime behaviour flags
 	syslogUseCurrentYear = flag.Bool("syslog_use_current_year", true, "Patch yearless timestamps with the present year.")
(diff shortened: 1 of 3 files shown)


cmd/mtail/main.go
```diff
@@ -47,6 +47,7 @@ var (
 	dumpAst      = flag.Bool("dump_ast", false, "Dump AST of programs after parse (to INFO log).")
 	dumpAstTypes = flag.Bool("dump_ast_types", false, "Dump AST of programs with type annotation after typecheck (to INFO log).")
 	dumpBytecode = flag.Bool("dump_bytecode", false, "Dump bytecode of programs (to INFO log).")
+	debug        = flag.Bool("debug", false, "Enable debugging endpoints.")
```
jaqx0r, line 50: Default to true to retain existing behaviour please.

jaqx0r: Also please make the flag be called "http_debugging_endpoint".

internal/mtail/mtail.go
```diff
@@ -79,6 +79,7 @@ type Server struct {
 	dumpAst      bool // if set, mtail prints the program syntax tree after parse
 	dumpAstTypes bool // if set, mtail prints the program syntax tree after type checking
 	dumpBytecode bool // if set, mtail prints the program bytecode after code generation
+	debug        bool // if set, enable debug endpoints
```
jaqx0r, line 82: This variable and it's friends should be called httpDebugEndpoints

internal/mtail/mtail.go
```diff
@@ -299,19 +300,23 @@ func (m *Server) Serve() error {
 		return errors.Errorf("No bind address provided.")
 	}
 	mux := http.NewServeMux()
+	if m.debug {
+		mux.Handle("/", m)
```
jaqx0r, line 304: I think the root handler should remain but we can strip the sensitive information in the template.

internal/mtail/options.go
```diff
@@ -107,6 +107,12 @@ func DumpBytecode(m *Server) error {
 	return nil
 }
 
+// Debug enables debug http endpoints
+func Debug(m *Server) error {
```
jaqx0r, line 111: EnableHTTPDebugEndpoints

PR #317, opened 2020-05-06
Switch mtail to use native go-fuzz for OSS-Fuzz

geethanjalieswaran:
Ref: https://github.com/google/oss-fuzz/pull/3638

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,8 @@ $(GOFUZZBUILD):
 
 GOFUZZ = $(GOBIN)/go-fuzz
 $(GOFUZZ):
-	go get $(GOGETFLAGS) github.com/dvyukov/go-fuzz/go-fuzz
+	go get $(GOGETFLAGS) github.com/mdempsky/go114-fuzz-build
+	ln -s $(GOPATH)/bin/go114-fuzz-build $(GOPATH)/bin/go-fuzz
 
 GOVERALLS = $(GOBIN)/goveralls
 $(GOVERALLS):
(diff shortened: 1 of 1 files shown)


geethanjalieswaran:
This change was tested on gcr.io/oss-fuzz-base/base-builder Docker image where https://github.com/mdempsky/go114-fuzz-build(native go-fuzz) is installed as bin/go-fuzz.

Since the CI for this repo uses gcr.io/fuzzit-public/buster-golang12 Docker image, the fuzzing check is failing at the moment. Also, to use github.com/mdempsky/go114-fuzz-build, go1.14 is required, so I guess this change has to be deferred until a fuzzit-public golang1.14 Docker image is released. 

One other option I see is to switch the CircusCI-fuzzing test to use gcr.io/oss-fuzz-base/base-builder, if that's suitable.

Makefile
```diff
@@ -236,8 +237,8 @@ CXXFLAGS ?=
 LIB_FUZZING_ENGINE ?= -libfuzzer
 OUT ?= .
 
-$(OUT)/vm-fuzzer: $(GOFILES) | $(GOFUZZBUILD)
-	$(GOFUZZBUILD) -libfuzzer -o fuzzer.a ./internal/vm
+$(OUT)/vm-fuzzer: $(GOFILES) | $(GOFUZZ)
+	$(GOFUZZ) -o fuzzer.a github.com/google/mtail/internal/vm
 	$(CXX) $(CXXFLAGS) $(LIB_FUZZING_ENGINE) fuzzer.a -lpthread -o $(OUT)/vm-fuzzer
```
jaqx0r, line 242: With the change above, you won't need to rename GOFUZZBUILD to GOFUZZ.

I am confused why this PR doesn't remove the CXX line though, if this change uses "native go-fuzz"; I don't see where go-fuzz is being invoked.

Makefile
```diff
@@ -72,7 +72,8 @@ $(GOFUZZBUILD):
 
 GOFUZZ = $(GOBIN)/go-fuzz
 $(GOFUZZ):
-	go get $(GOGETFLAGS) github.com/dvyukov/go-fuzz/go-fuzz
+	go get $(GOGETFLAGS) github.com/mdempsky/go114-fuzz-build
+	ln -s $(GOPATH)/bin/go114-fuzz-build $(GOPATH)/bin/go-fuzz
```
jaqx0r, line 76: This should have changed the GOFUZZBUILD rule above instead of this one, avoiding the symlink and the change of meaning.

jaqx0r:
going to drop this PR as I've fixed it elsewhere.

PR #318, opened 2020-05-20
added new examples/frequency_and_timer_record.mtail

AshishVarshneyy:
**SUMMARY**

This example help to understand the following things:
1. Read-only specific file otherwise halts the program.
2. We can assign values from log to any variable of mtail using the gauge.
3. We can use gauge and counter under single pattern match also.

**COMPONENT NAME**
examples/frequency_and_timer_record.mtail


AshishVarshneyy:
@googlebot I signed it!

AshishVarshneyy:
@googlebot I signed it!

PR #320, opened 2020-06-11
small comment fix in examples/apache_combined.mtail

sajoupa:
The 4th field is %t, not %u.


sajoupa:
@googlebot I signed it!

jaqx0r:
Thanks!

PR #325, opened 2020-06-25
Fix the bug that multi-level directory /a/*/c/*.log wildcard cannot be matched dynamically

young-0:
The main problems solved by this submission:
When the monitoring log file is --logs '/tmp/logs/*/log/*log'
There are currently /tmp/logs/a/log/a.log, /tmp/logs/b/log/b.log. No problem with monitoring.
When a new directory file /tmp/logs/c/log/c.log is added, it cannot be automatically added to the monitoring list.
Solution:
A new go routine has been added to regularly search for '/tmp/logs//log/*log' to join the monitoring list.


jaqx0r:
Excellent, thankyo!

PR #327, opened 2020-07-08
Adding UNIX Socket support option for HTTP endpoints

ArcticB:
Expose prometheus metrics through UNIX Socket endpoint is a crucial need for my infrastructure. Please consider this modification for the main project or add a similar option.


ArcticB:
Hello,
Did somebody took a decision on this proposition ? Using UNIX socket to serve metrics could be really interesting for preserving server side security for example to prevent an attacker crashing the service and then re-exposing a static view of the endpoint.
Thanks in advance

jnovack:
> preserving server side security for example to prevent an attacker crashing the service and then re-exposing a [contrived] view of the endpoint

Is that a thing people worry about??  Even if it is, the metrics dashboard is not your canary, your monitoring/orchestration layer should have caught the service crash/restart, and your security layer should've caught the new binary running.

The difference between opening `127.0.0.1:3903` or sending to `/var/run/mtail.sock` is just a different security mechanism, network vs file-system.  It's still only available locally, and only 1 process can serve it.

As a feature add, it's good, but it's novel at best.  I like it, it's fun and functional, but it is not in the name of "security".

SuperQ:
The better thing to do is add TLS and auth. For example, integrate [github.com/prometheus/exporter-toolkit/https](https://github.com/prometheus/exporter-toolkit/tree/master/https).

jaqx0r:
@ArcticB I agree with @jnovack that it seems cool but very niche to add this feature.  `mtail` is supposed to specialise in log parsing, not metric distribution.   Would the TLS option that @SuperQ has suggested be sufficient for your use case?

Your second comment suggests that you're worried about an attacker crashing mtail after already gaining access to the node, so I suspect that nothing we do would protect against that case -- security professionals like Kelly Shortridge recommend setting up nodes to be read-only and crash on change, for example in her Velocity Berlin keynote: https://www.oreilly.com/radar/controlled-chaos-the-inevitable-marriage-of-devops-and-security

ArcticB:
Hello,

Thank you for your replies, if I misspoke on server-side security, I apologize, I will try to be more clear on this point. We are using `mtail` to monitor log files on UNIX servers where multiple users have shell access as unprivileged users. It is quite difficult to restrict access on the `mtail` enpoint from local users & prevent them from "stealing" the port in case `mtail` process stops or crashes when exposing it over the _loopback_ interface.

I agree with you @SuperQ, adding TLS supports would be a good idea but it's not possible to do so now with our Prometheus configs. That's why adding UNIX socket supports would allow a good level of server-side isolation (we can set permissions on sockets) while keeping `mtail` lightweight (remote scraping could still be done by using ssh forwarding, it supports both UNIX sockets & tcp).

SuperQ:
Prometheus does not scrape over unix sockets.

If you want to avoid users from taking the mtail port binding, configure mtail to bind to a privileged port.

ArcticB:
Indeed, Prometheus does not scrape over UNIX sockets but nginx does, hence it's possible to "proxyfy" them on the _loopback_ of the Prometheus server.

Also, using privileged ports prevents normal users from running servers on them but it doesn't prevent those users from accessing them. In addition, it means that the `mtail` service need to be run as a privileged user (which is not the case for us).

Adding a TLS layer would fix this problem but as I said earlier it is not possible with our Prometheus environment.

jnovack:
As a manager for an IT department (*weird flex, I know*), I have an adage that I apply when someone suggests a workaround, fix or idea for an ongoing issue.  "Use technology to find people problems, but use people to solve people problems." 

What you have, is a classic "people problem".  You have semi-trusted people (you have given them accounts), who may, at some point go rogue or just want to investigate, even through curiosity.  By all means, use technology to find when they stray (`netstat` to check for port processes, `ps` to check for binary process ownership), but if you have a malicious user who is attempting to subvert another user's process and present false information, you have a personnel issue, and HR deals with personnel issues, not IT.

If this accounts are for non-employees, and this is some sort of tenet situation where non-trusted users can "buy accounts", then you have a design issue.

ArcticB:
Hardening a server and its various services is a good practice regardless of users being trusted enough or not. You don't go and tell people not to use e.g firewalls because if they're hacked they can go to the police...

But I think we've drifted way past the actual point of the ticket. UNIX domain sockets have been a thing since the very beginning of UNIX and are still widely used. It's not a revolutionary point of view that UNIX sockets are far easier to secure than TCP sockets (because they use file ACLs), and they can be used when TCP sockets are not available (like inside an unprivileged container with no network stack).

`Nginx`, `Varnish`, etc. all have UNIX socket support and this PR is ~20 lines. If my use case doesn't make sense to you I'll continue to use my own fork of `mtail`, but I do think this can be of interest to many Linux sysadmins.

cmd/mtail/main.go
```diff
@@ -36,6 +36,7 @@ var logs seqStringFlag
 var (
 	port               = flag.String("port", "3903", "HTTP port to listen on.")
 	address            = flag.String("address", "", "Host or IP address on which to bind HTTP listener")
+	unixSocket         = flag.String("unix_socket", "", "UNIX Socket to listen on.")
```
jnovack, line 39: I know that jax0r has a period in `port`, but other lines don't have periods to end the sentence.

jnovack (changes requested):
I am not an authoritative source.  These are suggestions.

cmd/mtail/main.go
jaqx0r: I think we should go with no fullstop in the flag descriptions so please drop it.   I'll take care of `port`.

jaqx0r:
OK I think you've convinced @jnovack that this is good enough and so I'll merge it, but I want you to write some integration tests in `internal/mtail/` (perhaps you call it `unix_socket_export_integration_test.go`) that start the server (see the use of `mtail.TestStartServer` in the other integration tests) and you should load a simple program (from `examples/` perhaps, `linecount.mtail` is a good one) and then write one line into the tailer (the `basic_tail_integration_test.go` is terribly simple to copy here) and then finally you need to scrape the socket and parse the content to perform the actual test.

You can copy some of the code from `internal/mtail/testing.go` that makes an HTTP client the old boring TCP way to fetch metrics from the TCP port.  You might make your own http.Transport and then supply that to something from the Prometheus libraries to use as the HTTP client that performs the scrape, and then query the Prometheus... thing.. (I don't know what you call it... Collector?) for the value you expected to collect.  Once that's all there, I'll believe this works, and I'll even make sure it keeps working after any future refactors.

Thanks!

internal/mtail/options.go
```diff
@@ -45,6 +45,16 @@ func BindAddress(address, port string) func(*Server) error {
 	}
 }
 
+// BindUnixSocket sets the UNIX socket path in Server.
+func BindUnixSocket(unixSocketPath string) func(*Server) error {
+	return func(m *Server) error {
+		m.bindUnixSocket = unixSocketPath
+		var err error
+		m.listener, err = net.Listen("unix", unixSocketPath)
```
jaqx0r, line 53: I'm actually shocked that this is all that is necessary to make this work.

jaqx0r (changes requested):
Per the comment thread, please supply a test that ensures this feature works.

ArcticB:
@googlebot I fixed it.

ArcticB:
@jaqx0r, I added a test named `unix_socket_export_integration_test.go`, it starts the `mtail` server on a unix socket with the program `linecount.mtail` and then creates an HTTP client to send a request (through UNIX socket) and validate the result. To do so, I reused the code from `basic_tail_integration_test.go`.

Ordinarily the socket is forwarded to the monitoring server with SSH tunneling and Prometheus accesses it locally through an proxy which make the link between the `mtail` UNIX Sockets and the Prometheus TCP connection. Therefore we don't need to test if the Prometheus can scrape mtail's output because we use the same HTTP handler.

Now about the CLA, I edited my branch directly on Github so the emails are wrong but I signed it with my account's username. Is that a problem or is it possible for you to fix it by hand ?

jaqx0r:
The last time I saw this happen with clabot I was told that the PR author has to connect with the correct account.  I just scanned the commits here and see you've got most with @ArcticB but one as lowercase 'a' arcticB and the avatar icon is just a grey octocat.  So I suspect the first commit you have is a different user.  Would you mind rebasing somehow, perhaps a squash, and doing a force push if you can't get that author to be able to sign with clabot?

Sorry for the trouble.

internal/mtail/unix_socket_export_integration_test.go
```diff
@@ -0,0 +1,185 @@
...
+		tb.Fatalf("%s: body was %s", err, buf.String())
+	}
+	glog.Infof("TestGetMetric: returned value for %s: %v", name, r[name])
+	return r[name]
+}
+
+
+// expectMetricDelta checks to see if the difference between a and b is want;
+// it assumes both values are float64s that came from a TestGetMetric.
+func expectMetricDelta(tb testing.TB, a, b interface{}, want float64) {
```
jaqx0r, line 128: I think you can reuse testing's ExpectMetricDelta instead of copying it here?

internal/mtail/unix_socket_export_integration_test.go
```diff
@@ -0,0 +1,185 @@
...
+		tb.Fatal(err)
+	}
+
+	return mtail.New(metrics.NewStore(), w, options...)
+}
+
+
+// startUNIXSocketServer creates a new Server serving through a UNIX
+// socket and starts it running. It returns the server, and a cleanup function.
+func startUNIXSocketServer(tb testing.TB, pollInterval time.Duration, enableFsNotify bool, options ...func(*mtail.Server) error) (*mtail.Server, func()) {
```
jaqx0r, line 42: Took me a while to realise why you can't reuse `TestStartServer()` but you should be able to reuse `TestMakeServer()` instead of adding your `makeServer()` above.

We could refactor `TestStartServer()` to notice if there's a BindUnixSocket optoin and change how the `net.DialTimeout` calls to wait for the server to start, if you like.  If not, please add a comment to the docstring explaining why this one is different.

internal/mtail/unix_socket_export_integration_test.go
```diff
@@ -0,0 +1,185 @@
...
+			tb.Fatal(err)
+		}
+	}
+}
+
+
+// getMetricFromUNIXSocket fetches the name metrics from the Server serving through
+// the UNIX socket sockPath, and returns the value of one named name.  Callers are
+// responsible for type assertions on the returned value.
+func getMetricFromUNIXSocket(tb testing.TB, sockPath, name string) interface{} {
```
jaqx0r, line 99: I think this should be a method on the *TestServer so that we can abstract away the `http.Transport`.  But all I'll ask is that you write that down as a TODO here.  Thanks!

internal/mtail/unix_socket_export_integration_test.go
```diff
@@ -0,0 +1,185 @@
...
+		b = 0.
+	}
+	delta := a.(float64) - b.(float64)
+	if delta != want {
+		tb.Errorf("Unexpected delta: got %v - %v = %g, want %g", a, b, delta, want)
+	}
+}
+
+
+func TestBasicUNIXSockets(t *testing.T) {
```
jaqx0r, line 143: :+1:

jaqx0r (changes requested):

internal/mtail/unix_socket_export_integration_test.go
ArcticB: Initially I tried to do it like you said, by using `mtail.ExpectMetricDelta` and `mtail.TestMakeServer` but when I tested it with the command `go test unix_socket_export_integration_test.go`, the methods were not defined. Perhaps I'm doing it the wrong way (I never used the Go test mechanism before and the Makefile targets don't execute the tests from the `mtail_test` package).

internal/mtail/unix_socket_export_integration_test.go
ArcticB: I will modify the docstring because refactor `TestStartServer` means that the other tests need also to be modified. It's a good idea but perhaps it need to be done later.

jaqx0r:
Ok no worries!

ArcticB:
I did a squash of my commits and pushed them on the my branch (from `git log`: _Author: ArcticB_).

jaqx0r (commented):
The rebase appears to have added commits, not squashed them, there's now 106 commits but I expected only 1 or at least all but the one that had the other author on it.  Can you please try it again?

jaqx0r:
https://github.com/google/mtail/pull/327/commits/522f47c3984bd228f8c7bbe098027c3237b27187 is the commit in question that needs to be squashed away.

ArcticB:
Here it is, sorry for the little inconvenience, I didn't drop the commits in my last squash attempt. It's good now.

jaqx0r (approved):

jaqx0r:
So close!  The branch is out of date; can you rebase it onto `master` please?

ArcticB:
I merged with the Github feature rather than rebase (so as not to fail the CLA again).

jaqx0r:
Thanks for all your patience!

PR #330, opened 2020-07-26
Makefile: Add linux/arm and linux/arm64 release

mrueg:
I'd like to run this on arm64 machines, can we have released binaries for it as well?

See also: 49d933dff0ed9cebb6f6a1f98bf63ade823a9953


mrueg:
@googlebot I signed it!

jaqx0r:
Thanks!

PR #459, opened 2021-01-27
fix SIGHUP handler: check for non-empty path to programs

veksh:
Current build is unable to reload programs on HUP, logging that

```
I0127 23:04:43.886579   13546 loader.go:434] no program reload on SIGHUP without programPath
```

apparently there is a trivial typo in path check.


veksh:
@googlebot I signed it!

jaqx0r:
Hah, nice.  Thankyou!

jaqx0r (approved):

PR #474, opened 2021-02-28
Added Configurable Semantic Checker Threshold Values

### LINKED ISSUE #471: compiler: Maximum recursion depth & maximum regular expression pattern length
Hi mTail team,

We are seeing the following errors upgrading from rc34 to anything above rc34:

`Expression exceeded maximum recursion depth.`

and 

`sipp.mtail:90:1-52: Exceeded maximum regular expression pattern length of 1024 bytes with 2820.`

(the later shows up when we resolve the former issue with a patch I'll describe later)

We have a fairly long mtail expression (couple hundred lines) to parse a fairly large CSV. We don't have control over the source program output so we can't really change the mtail expression. We also want to collect every single value so reducing what is captured by our regexs is not a solution for us. 

We found that the source of the problem comes from these lines in `checker.go`, presumably occuring during mtail  parsing:

`20: const kMaxRegexpLen = 1024`
`51:	if c.depth > 100 {`

Bumping these values up (to 3072 and 200 respectively) and rebuilding mtail locally seems to solve these problems.

Its my understanding these were added for performance reasons. However, in our specific situation, performance is not a concern for us as we use mtail for testing purposes. 

We're left with a dilemma here since, as far as I can tell, there is no way these values are configurable given they are hardcoded values. 

The only solution we can think of that both addresses the performance concerns and our requirements, without forking the repository for our own purposes, is to make these values configurable. I wanted to run it by you guys and understand if there is appetite for making these values configurable via the command line interface. Or configurable some other method would be acceptable too (environment variables?). If so, I'll be willing to submit a PR to implement that functionality (assuming I can figure out how to do it 👯 ).

Long story short, I'm asking, if I make these values configurable, will you be willing to merge my changes? 

Thanks,
Danielle

danielle-mustillo:
Previously, the semantic checker had static threshold values
that were added for performance reasons. However, some users
require higher threshold values and are willing to pay
the performance penalty.

This commit makes those values configurable but sets
defaults that are performant.

Fixes #471


internal/vm/checker/checker.go
```diff
@@ -27,16 +25,18 @@ type checker struct {
...
+	maxRegexLength    int
 }
 
 // Check performs a semantic check of the astNode, and returns a potentially
 // modified astNode and either a list of errors found, or nil if the program is
 // semantically valid.  At the completion of Check, the symbol table and type
 // annotation are also complete.
-func Check(node ast.Node) (ast.Node, error) {
-	c := &checker{}
+func Check(node ast.Node, maxRegexLength int, maxRecursionDepth int) (ast.Node, error) { // TODO is there a way to inject these integers without adding them as part of this constructor?
```
danielle-mustillo, line 38: Not sure if this is something you guys do, but I left this todo because I'm not very happy with the changes I had to make here. I want to make this better before merging.  But I can't think of any way to make it better. To any other reviewer, do you have a better idea of how I could inject these integers in?

jaqx0r:
Can you please sign the CLA?

danielle-mustillo:
@googlebot I signed it!

danielle-mustillo:
@googlebot I fixed it.

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	contrib.go.opencensus.io/exporter/jaeger v0.2.1
+	github.com/flazz/togo v0.0.0-20170320145504-babdbf21cff0 // indirect
```
jaqx0r, line 7: Revert this file

go.sum
```diff
@@ -132,6 +132,8 @@ github.com/envoyproxy/protoc-gen-validate v0.1.0 h1:EQciDnbrYxy13PgWoY8AqoxGiPrp
 github.com/envoyproxy/protoc-gen-validate v0.1.0/go.mod h1:iSmxcyjqTsJpI2R4NaDN7+kN2VEUnK/pcBlmesArF7c=
 github.com/fatih/color v1.7.0 h1:DkWD4oS2D8LGGgTQ6IvwJJXSL5Vp2ffcQg58nFV38Ys=
 github.com/fatih/color v1.7.0/go.mod h1:Zm6kSWBoL9eyXnKyktHP6abPY2pDugNf5KwzbycvMj4=
+github.com/flazz/togo v0.0.0-20170320145504-babdbf21cff0 h1:XWIs3kcxtU/euLH6qzv5jx3ixx81Sm0y7ARC2pf+lcA=
```
jaqx0r, line 135: Revrt this file

internal/mtail/examples_integration_test.go
```diff
@@ -92,7 +92,7 @@ func TestExamplePrograms(t *testing.T) {
 			waker, _ := waker.NewTest(ctx, 0) // oneshot means we should never need to wake the stream
 			store := metrics.NewStore()
 			programFile := filepath.Join("../..", tc.programfile)
-			mtail, err := mtail.New(ctx, store, mtail.ProgramPath(programFile), mtail.LogPathPatterns(tc.logfile), mtail.OneShot, mtail.OmitMetricSource, mtail.DumpAstTypes, mtail.DumpBytecode, mtail.LogPatternPollWaker(waker), mtail.LogstreamPollWaker(waker))
+			mtail, err := mtail.New(ctx, store, mtail.ProgramPath(programFile), mtail.LogPathPatterns(tc.logfile), mtail.OneShot, mtail.OmitMetricSource, mtail.DumpAstTypes, mtail.DumpBytecode, mtail.LogPatternPollWaker(waker), mtail.LogstreamPollWaker(waker), mtail.MaxRecursionDepth(100), mtail.MaxRegexLength(1024))
```
jaqx0r, line 95: If these options have defaults you shuoldnlt need to set them here.

internal/mtail/examples_integration_test.go
```diff
@@ -133,7 +133,7 @@ func TestCompileExamplePrograms(t *testing.T) {
 		t.Run(name, func(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			s := metrics.NewStore()
-			mtail, err := mtail.New(ctx, s, mtail.ProgramPath(tc), mtail.CompileOnly, mtail.OmitMetricSource, mtail.DumpAstTypes, mtail.DumpBytecode)
+			mtail, err := mtail.New(ctx, s, mtail.ProgramPath(tc), mtail.CompileOnly, mtail.OmitMetricSource, mtail.DumpAstTypes, mtail.DumpBytecode, mtail.MaxRegexLength(1024), mtail.MaxRecursionDepth(100))
```
jaqx0r, line 136: Ditto here

internal/mtail/examples_integration_test.go
```diff
@@ -218,15 +218,15 @@ func TestFilePipeStreamComparison(t *testing.T) {
 
 			go func() {
 				defer wg.Done()
-				fileMtail, err := mtail.New(ctx, fileStore, mtail.ProgramPath(programFile), mtail.LogPathPatterns(tc.logfile), mtail.OneShot, mtail.OmitMetricSource, mtail.LogPatternPollWaker(waker), mtail.LogstreamPollWaker(waker))
+				fileMtail, err := mtail.New(ctx, fileStore, mtail.ProgramPath(programFile), mtail.LogPathPatterns(tc.logfile), mtail.OneShot, mtail.OmitMetricSource, mtail.LogPatternPollWaker(waker), mtail.LogstreamPollWaker(waker), mtail.MaxRegexLength(1024), mtail.MaxRecursionDepth(100))
```
jaqx0r, line 221: etc

internal/mtail/examples_integration_test.go
```diff
@@ -218,15 +218,15 @@ func TestFilePipeStreamComparison(t *testing.T) {
...
 				if err != nil {
 					t.Error(err)
 				}
 				if err := fileMtail.Run(); err != nil {
 					t.Error(err)
 				}
 			}()
-			pipeMtail, err := mtail.New(ctx, pipeStore, mtail.ProgramPath(programFile), mtail.LogPathPatterns(pipeName), mtail.OneShot, mtail.OmitMetricSource, mtail.LogPatternPollWaker(waker), mtail.LogstreamPollWaker(waker))
+			pipeMtail, err := mtail.New(ctx, pipeStore, mtail.ProgramPath(programFile), mtail.LogPathPatterns(pipeName), mtail.OneShot, mtail.OmitMetricSource, mtail.LogPatternPollWaker(waker), mtail.LogstreamPollWaker(waker), mtail.MaxRegexLength(1024), mtail.MaxRecursionDepth(100))
```
jaqx0r, line 229: etc

internal/vm/checker/checker.go
jaqx0r: The checker and the rest of the compiler do not support the Options pattern, so this is your only choice right now, so drop the TODO.

internal/vm/checker/checker.go
```diff
@@ -17,8 +17,6 @@ import (
 	"github.com/google/mtail/internal/vm/types"
 )
 
-const kMaxRegexpLen = 1024
```
jaqx0r, line 20: Keep this but rename to 'kDefaultMaxRegexpLength' and add 'kDefaultMaxRecursionDepth', see below for why

internal/vm/checker/checker.go
```diff
@@ -27,16 +25,18 @@ type checker struct {
...
 }
 
 // Check performs a semantic check of the astNode, and returns a potentially
 // modified astNode and either a list of errors found, or nil if the program is
 // semantically valid.  At the completion of Check, the symbol table and type
 // annotation are also complete.
-func Check(node ast.Node) (ast.Node, error) {
-	c := &checker{}
+func Check(node ast.Node, maxRegexLength int, maxRecursionDepth int) (ast.Node, error) { // TODO is there a way to inject these integers without adding them as part of this constructor?
+	c := &checker{maxRegexLength: maxRegexLength, maxRecursionDepth: maxRecursionDepth}
```
jaqx0r, line 39: Prior to this line, set the defaults.    This means you won't have to pass them from the flags in the test consturctors.

if maxRegexpLength == 0 { maxRegexpLength = kDefaultMaxregexpLenght }
if maxrecursion... { }

internal/vm/checker/checker_test.go
```diff
@@ -262,7 +262,7 @@ func TestCheckInvalidPrograms(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			ast, err := parser.Parse(tc.name, strings.NewReader(tc.program))
 			testutil.FatalIfErr(t, err)
-			ast, err = checker.Check(ast)
+			ast, err = checker.Check(ast, 1024, 100)
```
jaqx0r, line 265: Wirth the zero => default above, you can change these params to 0.

internal/vm/compiler.go
```diff
@@ -17,7 +17,7 @@ import (
 // Compile compiles a program from the input into a virtual machine or a list
 // of compile errors.  It takes the program's name and the metric store as
 // additional arguments to build the virtual machine.
-func Compile(name string, input io.Reader, emitAst bool, emitAstTypes bool, syslogUseCurrentYear bool, loc *time.Location) (*VM, error) {
+func Compile(name string, input io.Reader, emitAst bool, emitAstTypes bool, syslogUseCurrentYear bool, loc *time.Location, maxRegexLength int, maxRecursionDepth int) (*VM, error) {
```
jaqx0r, line 20: This now looks like a prime candidate for Options pattern, I'll add that in a followup change.  Can you leave a TODO here?

internal/vm/loader.go
```diff
@@ -355,6 +357,22 @@ func SyslogUseCurrentYear() Option {
 	}
 }
 
+// MaxRegexLength sets the maximum length an mtail regular expression can have, in terms of characters.
```
jaqx0r, line 360: Regexp is the spelling used in `mtail`.

jaqx0r (changes requested):
This is good.  My comments only around passing params, defaults, and some spelling.

internal/mtail/examples_integration_test.go
danielle-mustillo: Unfortunately this is required for the unit tests pretty much across the board (ie. this file and etc).

The defaults are set in `cmd/main.go` via the `flag.Int` parameters (line 66 & 67 of that file in this merge request). In the production code, the `New` call on line 213 of `cmd/main.go` gets passed those flag values. Regardless whether it was set by the user or defaulted, the new call doens't know anything besides that it was passed a struct of `Option...` type. 

In this unit test, like the others, we have to take the role of passing those parameters. There is no `main.go` to do that for us since those tests don't have any interface to the command line (in the same way that main.go is a command line function). So hence we have all this duplication. 

I don't think this is the first time this has to be done. A similar example is the third argument here `mtail.ProgramPath(programFile)` is actually given a default on line 43 of `cmd/main.go` and passed into the new call on line 213 of that file. The unit tests don't invoke the main.go code so we have to duplicate that ProgramPath thing in a lot of tests.

This isn't great I agree, I'm happy to implement any suggestions to improve this if you have any. I'm not confident enough in this codebase to suggest one quite yet though!

danielle-mustillo: Never mind, I read your suggestion below. I agree I'll implement that.

cmd/mtail/main.go
```diff
@@ -63,6 +63,8 @@ var (
 	expiredMetricGcTickInterval = flag.Duration("expired_metrics_gc_interval", time.Hour, "interval between expired metric garbage collection runs")
 	staleLogGcTickInterval      = flag.Duration("stale_log_gc_interval", time.Hour, "interval between stale log garbage collection runs")
 	metricPushInterval          = flag.Duration("metric_push_interval", time.Minute, "interval between metric pushes to passive collectors")
+	maxRegexLength              = flag.Int("max_regex_length", 1024, "The maximum length a mtail regex expression can have. Excessively long patterns are likely to cause compilation and runtime performance problems.")
```
jaqx0r, line 66: max_regexp_length please

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/google/go-cmp v0.5.5
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.10.0
-	github.com/prometheus/common v0.19.0
+	github.com/prometheus/common v0.20.0
```
jaqx0r, line 12: revert this file please

go.sum
```diff
@@ -280,8 +280,8 @@ github.com/prometheus/common v0.4.1/go.mod h1:TNfzLD0ON7rHzMJeJkieUDPYmFC7Snx/y8
 github.com/prometheus/common v0.7.0/go.mod h1:DjGbpBbp5NYNiECxcL/VnbXCCaQpKd3tt26CguLLsqA=
 github.com/prometheus/common v0.10.0/go.mod h1:Tlit/dnDKsSWFlCLTWaA1cyBgKHSMdTB80sz/V91rCo=
 github.com/prometheus/common v0.18.0/go.mod h1:U+gB1OBLb1lF3O42bTCL+FK18tX9Oar16Clt/msog/s=
-github.com/prometheus/common v0.19.0 h1:Itb4+NjG9wRdkAWgVucbM/adyIXxEhbw0866e0uZE6A=
-github.com/prometheus/common v0.19.0/go.mod h1:U+gB1OBLb1lF3O42bTCL+FK18tX9Oar16Clt/msog/s=
+github.com/prometheus/common v0.20.0 h1:pfeDeUdQcIxOMutNjCejsEFp7qeP+/iltHSSmLpE+hU=
+github.com/prometheus/common v0.20.0/go.mod h1:U+gB1OBLb1lF3O42bTCL+FK18tX9Oar16Clt/msog/s=
```
jaqx0r, line 284: revert this file please

jaqx0r (changes requested):

cmd/mtail/main.go
danielle-mustillo: Good catch thanks!

jaqx0r (approved):
Thanks!  You just gotta rebase or sync and then all ready to go.

jaqx0r:
Sorry looks like it's all ready, no need to sync.

PR #478, opened 2021-03-09
Improve performance of finding LabelValue

saddays:
Sorry, i fouled up the last [PR](https://github.com/google/mtail/pull/477) , please review this.

----------------------------
I use mtail to calulate nginx logs, the script looks like this:
```
histogram  req  buckets 0,10,100,1000 by host,request_uri, upstream_addr, status_code
```

When logs exceed 400K/per minute And label combinations exceed 10K, the mtail seems to get stuck (keep cpu_usage 110% and cann't handle the current minute logs). After profiling and reading the source code ,I found the time complexity of finding labelvalue is O(M*N), where M is the count of LabelValue, and N is the count of LabelKey.

source file mtail/internal/metrics/metric.go
```
func (m *Metric) FindLabelValueOrNil(labelvalues []string) *LabelValue {
Loop:
	for i, lv := range m.LabelValues {
		for j := 0; j < len(lv.Labels); j++ {
			if lv.Labels[j] != labelvalues[j] {
				continue Loop
			}
		}
		return m.LabelValues[i]
	}
	return nil
}
```

After optimizing, it looks to be 5 times faster than rc44 in my use case. Please help me review it , thx.


saddays:
@jaqx0r    i fouled up the last PR , please review this.

jaqx0r (changes requested):
Can you add a benchmark test please?

saddays:
Test updated :)

On my local laptop, the boundary is single metric has more than 500 labelvalue with 4 labels.  While exceed 500, the patch version is more efficient.

jaqx0r:
The test failure is a known flake.  Looks good!

jaqx0r (dismissed):

internal/metrics/metric.go
```diff
@@ -147,7 +157,10 @@ func (m *Metric) GetDatum(labelvalues ...string) (d datum.Datum, err error) {
 			}
 			d = datum.NewBuckets(buckets)
 		}
-		m.LabelValues = append(m.LabelValues, &LabelValue{Labels: labelvalues, Value: d})
+		lv := &LabelValue{Labels: labelvalues, Value: d}
+		m.LabelValues = append(m.LabelValues, lv)
```
jaqx0r, line 161: Do we need both the map and the list now?

saddays: Actuactually not need,  it's just for compatible. Drop the list maybe improve the performance further.

jaqx0r (approved):

PR #501, opened 2021-04-09
typo in line 124

georgecma:
line 124 subst function has no matching right parenthesis.


jaqx0r:
This got fixed in #500

PR #575, opened 2021-08-20
examples/apache_metrics.mtail: fix http_request_duration_seconds division

mika:
We need to use floating point division, instead of integer division,
otherwise the values end up as wrong and unexpected numbers in the
histogram's buckets.


mika:
@googlebot I signed it!

jaqx0r:
Thanks!

jaqx0r (approved):

PR #576, opened 2021-08-23
Exit with 0 code when --one_shot is set

prymitive:
I was able to use --one_shot to test mtail file syntax but ab3cbf22d2359f9993c5885ae19a11111ee97c0b added os.Exit(1) so I cannot use exit code to tell if a file is valid or not.
Change exit code from 1 to 0, if there was a syntax error then earlier code will catch it.


prymitive:
@googlebot I signed it!

jaqx0r:
Nice, I think the exit 1 should be inside the if statement just a few lines above. Can you add one there to exit 1 if the error is logged?

PR #603, opened 2021-11-24
Split pollInterval option to reduce filepath.Glob overhead

Loyilee:
Hi,  i'm using mtail to monitor the docker logs of the machine, and every machine has 500+ matched log files .  Below is the configuration:

```
mtail --progs /etc/mtail --logs "/home/docker/logs/*/*/*/*/*/platform-*.log" --expired_metrics_gc_interval 10m
```

I notice that mtail always keeps more than 60% cpu usage, **even no logs produced**.   After profiling the process, i found that the primary reason is **tailer#PollLogPatterns**  calls frequently (run per 250ms by default).  We can't separately control the interval of poll-log-pattern or poll-log-stream.  I tried to split the option and deployed the fixed mtail on production for a few weeks, and it works well.  Could you help me to review the feature?

### DIFF
--- a/cmd/mtail/main.go
+++ b/cmd/mtail/main.go
@@ -62,7 +62,8 @@ var (
 	logRuntimeErrors     = flag.Bool("vm_logs_runtime_errors", true, "Enables logging of runtime errors to the standard log.  Set to false to only have the errors printed to the HTTP console.")
 
 	// Ops flags.
-	pollInterval                = flag.Duration("poll_interval", 250*time.Millisecond, "Set the interval to poll all log files for data; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
+	pollInterval                = flag.Duration("poll_interval", 250*time.Millisecond, "Set the interval to poll all log files for list; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
+	pollLogInterval             = flag.Duration("poll_log_interval", 250*time.Millisecond, "Set the interval to poll all log files for data; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
 	expiredMetricGcTickInterval = flag.Duration("expired_metrics_gc_interval", time.Hour, "interval between expired metric garbage collection runs")
 	staleLogGcTickInterval      = flag.Duration("stale_log_gc_interval", time.Hour, "interval between stale log garbage collection runs")
 	metricPushInterval          = flag.Duration("metric_push_interval", time.Minute, "interval between metric pushes to passive collectors")
@@ -147,6 +148,10 @@ func main() {
 		glog.Infof("no poll interval specified; defaulting to 250ms poll")
 		*pollInterval = time.Millisecond * 250
 	}
+	if *pollLogInterval == 0 {
+		glog.Infof("no poll interval specified; defaulting to pollInterval")
+		*pollLogInterval = *pollInterval
+	}
 
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
@@ -179,7 +184,8 @@ func main() {
 	}
 	if *pollInterval > 0 {
 		logPatternPollWaker := waker.NewTimed(ctx, *pollInterval)
-		opts = append(opts, mtail.LogPatternPollWaker(logPatternPollWaker), mtail.LogstreamPollWaker(logPatternPollWaker))
+		logStreamPollWaker := waker.NewTimed(ctx, *pollLogInterval)
+		opts = append(opts, mtail.LogPatternPollWaker(logPatternPollWaker), mtail.LogstreamPollWaker(logStreamPollWaker))
 	}
 	if *unixSocket == "" {
 		opts = append(opts, mtail.BindAddress(*address, *port))
--- a/cmd/mtail/main.go
+++ b/cmd/mtail/main.go
@@ -62,8 +62,8 @@ var (
 	logRuntimeErrors     = flag.Bool("vm_logs_runtime_errors", true, "Enables logging of runtime errors to the standard log.  Set to false to only have the errors printed to the HTTP console.")
 
 	// Ops flags.
-	pollInterval                = flag.Duration("poll_interval", 250*time.Millisecond, "Set the interval to poll all log files for list; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
-	pollLogInterval             = flag.Duration("poll_log_interval", 250*time.Millisecond, "Set the interval to poll all log files for data; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
+	pollInterval                = flag.Duration("poll_interval", 250*time.Millisecond, "Set the interval to poll each log file for data; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
+	pollLogInterval             = flag.Duration("poll_log_interval", 250*time.Millisecond, "Set the interval to find all matched log files for polling; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
 	expiredMetricGcTickInterval = flag.Duration("expired_metrics_gc_interval", time.Hour, "interval between expired metric garbage collection runs")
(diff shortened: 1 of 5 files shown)


jaqx0r:
Thanks for the PR!  The code looks reasonable but I have some more questions first.

60% is unexpectedly high.  I moved from a fsnotify to a poll model to *reduce* the CPU load!    I'm assuming though that calling glob to look for new log files every 250ms is actually unnecessary.

Can you tell me more about the problem?  How many logs are under mtail's watch?  How often do they get updates?

When you created this change, what did you set your new flag to -- how often do you do the log pattern poll?

Loyilee:
I have addressd out your comments, please take a look.

> Can you tell me more about the problem? How many logs are under mtail's watch? How often do they get updates?

1.  How many logs are under mtail's watch? How often do they get updates?
   
``` 
There are more than 2000 logs under the log path (/home/docker/logs/), about 400 logs match the pattern, but only 50+ logs are active updated (append logs per second),  others never appended but still retained for 1 week. 
```

2.  what did you set your new flag to?  how often do you do the log pattern poll?

``` 
It's ok for us just poll the pattern once every minute, and i don't  change the interval of polling logs (250ms).  

-expired_metrics_gc_interval 10m  -poll_interval 1m   -poll_log_interval 250ms
``` 

3. other supplements

Here is the result of  **top** shows, no logs are appending at this moment.
image

I tried to run **pprof**  30 seconds, here is the top100.
image

jaqx0r:
Great, thanks for the detailed information!  When you poll the glob at 1m, what happens the CPU in top?

cmd/mtail/main.go
```diff
@@ -62,7 +62,8 @@ var (
 	logRuntimeErrors     = flag.Bool("vm_logs_runtime_errors", true, "Enables logging of runtime errors to the standard log.  Set to false to only have the errors printed to the HTTP console.")
 
 	// Ops flags.
-	pollInterval                = flag.Duration("poll_interval", 250*time.Millisecond, "Set the interval to poll all log files for data; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
+	pollInterval                = flag.Duration("poll_interval", 250*time.Millisecond, "Set the interval to poll all log files for list; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
+	pollLogInterval             = flag.Duration("poll_log_interval", 250*time.Millisecond, "Set the interval to poll all log files for data; must be positive, or zero to disable polling.  With polling mode, only the files found at mtail startup will be polled.")
```
jaqx0r, line 66: I'd like these inverted; the original flag should remain for the log stream poll interval, and the new flag for the pattern/list interval please.

The help string for both need to be updated as well.

cmd/mtail/main.go
```diff
@@ -147,6 +148,10 @@ func main() {
 		glog.Infof("no poll interval specified; defaulting to 250ms poll")
 		*pollInterval = time.Millisecond * 250
 	}
+	if *pollLogInterval == 0 {
+		glog.Infof("no poll interval specified; defaulting to pollInterval")
```
jaqx0r, line 152: Once this is changed to be the pattern interval, then this and line 148 should explain which poll interval is being warned about.

cmd/mtail/main.go
```diff
@@ -179,7 +184,8 @@ func main() {
 	}
 	if *pollInterval > 0 {
 		logPatternPollWaker := waker.NewTimed(ctx, *pollInterval)
-		opts = append(opts, mtail.LogPatternPollWaker(logPatternPollWaker), mtail.LogstreamPollWaker(logPatternPollWaker))
+		logStreamPollWaker := waker.NewTimed(ctx, *pollLogInterval)
```
jaqx0r, line 187: as above, please invert

jaqx0r (changes requested):
I'd also like the new flag mentioned in docs/Deployment.md please.

Loyilee:
> Great, thanks for the detailed information! When you poll the glob at 1m, what happens the CPU in top?

Normally less than 6%

image

cmd/mtail/main.go
Loyilee: ok, i'd invert them

cmd/mtail/main.go
Loyilee: I have updated the description, please take a look.

jaqx0r (changes requested):
Thanks!

jaqx0r (approved):
I pressed the wrong button.  This time for sure.

jaqx0r:
There's been some flakes which I think are because the tests are now missing a construction option after the split here.

e.g. `internal/tailer/tail_test.go` line 27 the tailer is contstructed with only the LogstreamPollWaker, and I think that's causing the first set of writes in TestHandleLogTruncate to be missed in https://github.com/google/mtail/runs/4322005799?check_suite_focus=true

Likewise in `internal/mtail/testing.go` line 57, which is also missing the new option, which I think is causing the flake at https://github.com/google/mtail/runs/4320603579?check_suite_focus=true

jaqx0r (changes requested):
Per last comment, internal/mtail/testing.go and internal/tailer/tail_test.go need the new option in the constructor.

Loyilee:
@jaqx0r  I think it may be a unstable test case, not relevant with the new feature.  On my  laptop, the test failure is not always occurs. After debuging, i think it maybe caused by the filesystem flush buffers, the **double sync** in **TestHandleLogTruncate** seems not works.   

```
func TestHandleLogTruncate(t *testing.T) {	
       .........
       .........

	testutil.WriteString(t, f, "a\nb\nc\n")
	awaken(1)
	awaken(1) // double sync to ensure that a whole pass through the filestream has occurred

	if err := f.Truncate(0); err != nil {
		t.Fatal(err)
	}
	// "File.Truncate" does not change the file offset, force a seek to start.
	_, err := f.Seek(0, 0)
	testutil.FatalIfErr(t, err)
	awaken(1)

        // print the length of channel after truncate
	println(len(lines))
	println("###")
        
        .......
        .......
}
```

On the failure case, the length of lines is always **0** , that means we have turncate the log before we read the lines.

image

Loyilee:
Also i found some failure test case drop the last 2 lines.

jaqx0r:
It may well be, but these tests had no flakes before (I'd use the `-count 1000` flag to shake them out) so this change has caused them to make whatever latent problem visible again.

the `awaken(1)` call signals the testWaker to block until 1 routine is waiting, and then wake it.  I suppose there could still be a race that the WriteString above hasn't flushed to disk?  Can you insert a f.Sync() just before the awaken calls?

jaqx0r:
I've made my own changes to force a Sync after WriteString when the test is a regular file, which means I can remove that double-awaken.  You're right, it was still racy, because the data wasn't guaranteed to be where I thought it was.  Thanks for discovering that!

Loyilee:
OK,  i will rebase the code once your pr merged.

jaqx0r:
#604 merged!

jaqx0r (approved):

Loyilee:
All checks passed.

jaqx0r:
Thanks!

jaqx0r:
I've figured out the cause of the test failure.

Previously the function makeTestTail in test_tail.go did not create a
goroutine that called PollLogPatterns, we relied on the awaken() method
returned from the testWaker to wake the filestream.

With this change, the goroutine is started in the option for both wakers,
which means we now have two sources of a wake being triggered, which means
we can't control the sequencing of the test to ensure the data written
before the truncate is also read before the truncate.  The second ghost
waker steals the slot from the waker in test, so the function returns too
soon.

No big deal but a bit tricky to debug.  I'll figure out how to unwind the
goroutine from the option handler.

jaqx0r:
Aha it's even easier.  I forgot this during review, but I was feeling weird
about the addition of the StartLogStreamPollLoop method.  Now I understand
why.

We don't need that as the logstreamwaker is passed to the streams. In this
PR the waker is reused for the stream GC.  I think the stream GC should
stay using the pattern waker.

Loyilee:
You are right, i'm sorry to introduce that.   It shouldn't need so many changes.  Should i open another PR to fix it ?

jaqx0r:
No all good, I've just finished the change.  Thanks for offering!

Issue #611, opened 2021-12-31
Frequent test failures due to timeouts


NightTsarina:
Hi,

This is a problem that is affecting the Debian package for mtail, but I can see it affecting anybody contributing to mtail while using a mid-range computer. Some of the tests take too long to complete, I get consistent test errors while running them in my reasonably fast laptop if there are other processes (like a browser) using a non-trivial amount of CPU.

After locally increasing the test timeouts, these are the longest-running tests (in a pristine clone of this repo, no Debian patches):

```
--- PASS: TestExamplePrograms (4.01s)
    --- PASS: TestExamplePrograms/examples/dhcpd.mtail_on_testdata/anonymised_dhcpd_log (3.81s)
--- PASS: TestFilePipeStreamComparison (3.03s)
    --- PASS: TestFilePipeStreamComparison/examples/dhcpd.mtail_on_testdata/anonymised_dhcpd_log (2.89s)
--- PASS: TestFileSocketStreamComparison (6.59s)
    --- PASS: TestFileSocketStreamComparison/examples/dhcpd.mtail_on_unixgram://testdata/anonymised_dhcpd_log (3.46s)
    --- PASS: TestFileSocketStreamComparison/examples/dhcpd.mtail_on_unix://testdata/anonymised_dhcpd_log (2.83s)
--- PASS: TestInsertLookupQuick (2.61s)
```

I will follow-up with a PR to increase these timeouts.

NightTsarina:
For race tests, this is even more pronounced:
```
--- PASS: TestExamplePrograms (40.16s)
    --- PASS: TestExamplePrograms/examples/dhcpd.mtail_on_testdata/anonymised_dhcpd_log (39.06s)
--- PASS: TestFilePipeStreamComparison (52.78s)
--- PASS: TestFileSocketStreamComparison (113.16s)
    --- PASS: TestFileSocketStreamComparison/examples/dhcpd.mtail_on_unixgram://testdata/anonymise
d_dhcpd_log (56.81s)
    --- PASS: TestFileSocketStreamComparison/examples/dhcpd.mtail_on_unix://testdata/anonymised_dh
cpd_log (54.67s)
--- PASS: TestInsertLookupQuick (12.75s)
```

jaqx0r:
Thanks.  I am wary of increasing the timeouts in case they mask other problems but I also don't have a good sense of how fast they should take.  Also I am pretty sure go treats the timeout for the entire suite and not per test which is frustrating for being able to tune them.

PR #612, opened 2021-12-31
Increase test timeouts

NightTsarina:
This PR increases the timeouts for tests using TestTimeout, as well as the `go
test` timeout, to avoid frequent failures in slow machines.

Closes: #611


jaqx0r (approved):

NightTsarina:
Thanks for the approval, but it seems it won't let me merge anyway.

jaqx0r:
Oh perhaps that is my fault. Done

Issue #609, opened 2021-12-31
Mtail dies if a log file disappears during initialisation


NightTsarina:
Hi!

I am running mtail to tail a big number of logs in a bunch of servers, which get turned off when they have no work to do. I have noticed that sometimes when these machines are powered back on, mtail fails to start:

```
Dec 31 00:34:55 bp16 sh[753]: I1231 00:34:55.116385     753 main.go:115] mtail version 3.0.0~rc48 git revision 3.0.0~rc48-1~bpo11+1 go version go1.15.15 go arch amd64 go os linux
Dec 31 00:34:55 bp16 sh[753]: I1231 00:34:55.116425     753 main.go:116] Commandline: ["/usr/bin/mtail" "--progs" "/etc/mtail" "--logtostderr" "--address" "::" "--port" "3903" "--logs" "/var/repositories/*/logs/*.log"]
Dec 31 00:34:55 bp16 sh[753]: I1231 00:34:55.116655     753 store.go:182] Starting metric store expiry loop every 1h0m0s

[..]

Dec 31 00:34:55 bp16 sh[753]: I1231 00:34:55.192237     753 logstream.go:61] Parsed url as /var/repositories/XXX/logs/YYY-1-2021-12-30.log
Dec 31 00:34:55 bp16 sh[753]: I1231 00:34:55.192259     753 tail.go:282] Tailing /var/repositories/XXX/logs/YYY-1-2021-12-30.log
Dec 31 00:34:55 bp16 sh[753]: E1231 00:34:55.192275     753 main.go:220] stat /var/repositories/XXX/logs/YYY-10-2021-12-26.log: no such file or directory
Dec 31 00:34:55 bp16 systemd[1]: mtail.service: Service hold-off time over, scheduling restart.
```

A similar message is logged when logrotate runs, but that does not make mtail crash; it just ignores the file and continues:
```
Dec 31 00:34:54 bp16 sh[643]: I1231 00:34:54.753486     643 filestream.go:139] stat /var/repositories/XXX/logs/YYY-84-2021-12-26.log: no such file or directory
```

My theory is that this happens when mtail is running `tailer.New` at the same time as logrotate is working (which can take many seconds to finish) during boot. Sadly, it is difficult to reproduce the exact same scenario, but the same behaviour happens if a create a broken link matching the log pattern:

```
$ ln -s /nonexistant /home/tinafoo.log; mtail -progs /etc/mtail -logs /home/tina/\*.log -port 9999 -logtostderr
[...]

PR #613, opened 2022-01-03
Skip logfiles that cannot be read

NightTsarina:
The Tailer.PollLogPatterns and Tailer.Ignore methods currently abort they
encounter files that cause errors on os.Stat or os.Getwd, which can be caused
by race conditions (file removed after filepath.Glob was called) or broken
symlinks. This can make mtail crash on start-up, but also prevents good log
files from being processed due to the early return of PollLogPatterns.

This commit changes PollLogPatterns to log the errors and continue the
operation. In Ignore the errors now are handled by returning `true`, so the
file is ignored.

Closes: #609

### DIFF
--- a/internal/tailer/tail.go
+++ b/internal/tailer/tail.go
@@ -227,20 +227,22 @@ func (t *Tailer) AddPattern(pattern string) error {
 	return nil
 }
 
-func (t *Tailer) Ignore(pathname string) (bool, error) {
+func (t *Tailer) Ignore(pathname string) bool {
 	absPath, err := filepath.Abs(pathname)
 	if err != nil {
-		return false, err
+		glog.V(2).Infof("Couldn't get absolute path for %q: %s", pathname, err)
+		return true
 	}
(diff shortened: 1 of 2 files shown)


internal/tailer/tail.go
```diff
@@ -362,16 +364,14 @@ func (t *Tailer) PollLogPatterns() error {
 		}
 		glog.V(1).Infof("glob matches: %v", matches)
 		for _, pathname := range matches {
-			ignore, err := t.Ignore(pathname)
-			if err != nil {
-				return err
-			}
+			ignore := t.Ignore(pathname)
```
jaqx0r, line 367: Collapse this into the following if statement:
```
if t.Ignore(pathname) {
  continue
}
```

jaqx0r (changes requested):

NightTsarina:
Comment addressed. PTAL?

jaqx0r (approved):

jaqx0r:
Thanks!

Issue #572, opened 2021-08-09
Accepting windows disk path for logs argument


TheArturthur:
Hello there! As specified in the guide, the issue I'm encountering is the following:

## Mtail version:
`mtail version 3.0.0-rc47 git revision 5e0099f843e4e4f2b7189c21019de18eb49181bf go version go1.16.5 go arch amd64 go os windows`

## Command run:
`PS C:\Exporters\Mtail_Exporter> .\mtail-exporter.exe -progs="C:\Exporters\Mtail_Exporter\Files" -logs="E:\SQLSERVER\full_diario.log"`

## Error returned:
`E0809 11:42:16.507262    6176 main.go:219] unsupported URL scheme: "e" in path pattern "E:\\SQLSERVER\\full_diario.log"`

## Problem:
It seems like the program doesn't accept specifying a Windows disk letter. I've tried different ways, but it appears as if it only accepts UNIX paths or Windows paths in the same drive.

Is there a workaround? Am I missing something?

jaqx0r:
Thanks for the detailed report!  You're not missing anything -- `mtail` is developed on Linux and so the windows support is lacking and relies on users like you to tell me where I messed up.

TheArturthur:
No problem! Glad to help! I also noticed that the labelling as explained here [Metrics/Labelling](https://github.com/google/mtail/blob/main/docs/Metrics.md#labelling), can't be done, as it doesn't show the metric at all. In Linux, I've tried it and works correctly. Should I create another issue for that?

aRestless:
The original issue can be closed due to https://github.com/google/mtail/pull/620

The metric labeling not working is something I'm not aware of on Windows (and I use mtail on Windows), so I guess a new issue with a reproduction case would be warranted.

lindhor:
Thanks for this, it saved my day to find this fix!

PR #620, opened 2022-01-31
Fix Windows path handling and all tests on Windows

aRestless:
This PR primarily addresses #572, but also fixes all other tests on Windows. Summary of the important changes:

- In logstream.go and tail.go, when the URL protocol cannot be recognized, the whole argument is now treated as a path. That leads to supporting file paths like `C:\somefile`. It should be noted though that local file paths and URLs are two separate concepts that are incompatible in edge cases, even on Linux, since colons are valid characters for file names. So it might be worth it to separate these two use cases in the future.
- File handles in tests are now explicitly closed, otherwise the temp folder cleanup in fs.go will not be able to delete the files on Windows
- All tests regarding pipes and sockets have been moved to Unix-only tests. It might be possible to get at least some of them running on Windows in the future.
- All tests relying on "permission denied" errors have been moved to Unix-only tests. On Windows, it is not possible to create a file that you cannot read yourself (minimum file permissions are 0222).


internal/mtail/multiple_levels_directory_integration_test.go
```diff
@@ -4,13 +4,11 @@
 package mtail_test
 
 import (
+	"github.com/google/mtail/internal/mtail"
+	"github.com/google/mtail/internal/testutil"
```
jaqx0r, line 8: Put the local imports last please.

internal/mtail/multiple_levels_directory_integration_test.go
```diff
@@ -28,9 +26,10 @@ func TestPollLogPathPatterns(t *testing.T) {
 	lineCountCheck := m.ExpectExpvarDeltaWithDeadline("lines_total", 1)
 
 	logFile := filepath.Join(logDir, "files", "a", "log", "a.log")
-	testutil.FatalIfErr(t, os.MkdirAll(path.Dir(logFile), 0700))
+	testutil.FatalIfErr(t, os.MkdirAll(filepath.Dir(logFile), 0700))
 
 	f := testutil.TestOpenFile(t, logFile)
+	defer f.Close()
```
jaqx0r, line 32: I think TestOpenFile should register a t.Cleanup() function to handle the Close instead of deferring in test here, but I'll take care of that in another PR.

internal/mtail/permission_denied_integration_unix_test.go
```diff
@@ -12,6 +15,8 @@ import (
 	"github.com/google/mtail/internal/testutil"
 )
 
+// TestPermissionDeniedOnLog is a unix-specific test because on Windows, it is not possible to create a file
+// that you yourself cannot read (minimum permissions are 0222).
```
jaqx0r, line 19: Thanks for documenting that!

jaqx0r (changes requested):
Thanks for the contribution!

Thanks for documenting why the tests are unix specific.

I have a question about log rotation then -- on Windows is the standard practice to truncate the file?  Or do logs never get rotated?  A file can't be renamed if it's open?

I didn't completely check but it looks like the LogTruncation test just happens to test the Windows behaviour.  Is there any other way that Windows log files are handled that we should cover?  I just feel that the integration tests have reduced coverage.

I'm not sure if you also want to make sure the end to end tests run on Windows, just to make sure that we cover any log-tailing bugs -- this test has helped me find errors connecting the tailer to the VM in the past.  

The only other thing is that the PR needs rebasing or merging with master, and the lint errors need addressing.

Thanks again!

internal/mtail/multiple_levels_directory_integration_test.go
aRestless: Very good idea!

aRestless:
> I have a question about log rotation then -- on Windows is the standard practice to truncate the file? Or do logs never get rotated? A file can't be renamed if it's open?

I also don't see myself as an expert but the practice I've seen is that indeed, the current log file is being copied and then truncated. In a similar manner, on Windows you also can't truncate a file that you've opened in append mode. That is why I switched from `TestOpenFile` to `OpenLogFile` in `TestHandleLogTruncate`.

> I didn't completely check but it looks like the LogTruncation test just happens to test the Windows behaviour. Is there any other way that Windows log files are handled that we should cover? I just feel that the integration tests have reduced coverage.

I would agree that the LogTruncation tests cover the Windows use case. The only other oddity I know of are the FILE_SHARE_READ and FILE_SHARE_WRITE "share mode" flags when opening files on Windows. They allow other processes to still access the file. Testing access to files opened without those flags could in theory be a substitute for the tests relying on permission denied errors for Linux. However, Golang _always_ sets those flags, so it's not possible to simulate from within a Golang test.

> I'm not sure if you also want to make sure the end to end tests run on Windows, just to make sure that we cover any log-tailing bugs -- this test has helped me find errors connecting the tailer to the VM in the past.

I feel immensely blind but can you point me towards where the end to end tests are defined? But regardless, that might be something for a future PR.

> The only other thing is that the PR needs rebasing or merging with master, and the lint errors need addressing.

Will get on that now. Thanks!

aRestless:
Linting errors should be fixed. I shot gofumpt/gofmt against everything and that has turned all octal notations to the form 0o666. Please let me know if that's alright or if that should be reverted.

Once all checks are green I'd rebase and squash for an easy merge and a clean history.

internal/mtail/multiple_levels_directory_integration_test.go
aRestless: done

jaqx0r:
> > I have a question about log rotation then -- on Windows is the standard practice to truncate the file? Or do logs never get rotated? A file can't be renamed if it's open?
> 
> I also don't see myself as an expert but the practice I've seen is that indeed, the current log file is being copied and then truncated. In a similar manner, on Windows you also can't truncate a file that you've opened in append mode. That is why I switched from `TestOpenFile` to `OpenLogFile` in `TestHandleLogTruncate`.

Ok, so in Windows, a process's log library would handle copying its own log file, and then doing the truncation.  Or is it an external process that would do this?

When you say you've seen this practice, can you give me an example of an application that does this?

> > I didn't completely check but it looks like the LogTruncation test just happens to test the Windows behaviour. Is there any other way that Windows log files are handled that we should cover? I just feel that the integration tests have reduced coverage.
> 
> I would agree that the LogTruncation tests cover the Windows use case. The only other oddity I know of are the FILE_SHARE_READ and FILE_SHARE_WRITE "share mode" flags when opening files on Windows. They allow other processes to still access the file. Testing access to files opened without those flags could in theory be a substitute for the tests relying on permission denied errors for Linux. However, Golang _always_ sets those flags, so it's not possible to simulate from within a Golang test.

Intriguing.  Thanks for the details.  We can come back to this comment thread in the future for reference.  (I'm wondering about skipping the file abstraction library and going straight into Windows syscalls if we need to.)

I'm satisfied with not having the permission denied test in Windows as it was an edge case that showed up during deployment of a very early version of `mtail`.

I think my feeling of coverage is entirely because I don't understand file semantics on windows and I don't know how log rotation works there either :)


 
> > I'm not sure if you also want to make sure the end to end tests run on Windows, just to make sure that we cover any log-tailing bugs -- this test has helped me find errors connecting the tailer to the VM in the past.
> 
> I feel immensely blind but can you point me towards where the end to end tests are defined? But regardless, that might be something for a future PR.

One of the tests you edited is called examples_integration_test, which you've made a `+build linux` only test.  If you can replicate the file-read part, the rest should be identical between linux and windows versions.  Happy to see it in a followup.

internal/mtail/basic_tail_integration_test.go
```diff
@@ -4,52 +4,14 @@
...
 	"os"
 	"path/filepath"
-	"sync"
 	"testing"
 
 	"github.com/google/mtail/internal/mtail"
 	"github.com/google/mtail/internal/testutil"
 )
 
-func TestBasicTail(t *testing.T) {
```
jaqx0r, line 17: If this test can be made to work on Windows, that would be ideal.  It really just checks that the tailer can open a file, which I expect should also work on Windows.

internal/mtail/examples_integration_unix_test.go
```diff
@@ -0,0 +1,172 @@
+// Copyright 2011 Google Inc. All Rights Reserved.
+// This file is available under the Apache license.
+
+//go:build unix
+// +build unix
+
+package mtail_test
+
+// Two mtails both alike in dignity.
```
jaqx0r, line 9: Pop a note here like you did for the other tests saying that Pipes are unix only thing.

internal/mtail/golden/reader_test.go
```diff
@@ -106,7 +106,7 @@ var expectedMetrics = metrics.MetricSlice{
 		LabelValues: []*metrics.LabelValue{
 			{
 				Labels: []string{},
-				Value:  datum.MakeString("hi", time.Date(2018, 6, 16, 18, 04, 0, 0, time.UTC)),
+				Value:  datum.MakeString("hi", time.Date(2018, 6, 16, 18, 0o4, 0, 0, time.UTC)),
```
jaqx0r, line 109: This one's a bug!  Just make it a decimal 4 please.

internal/runtime/vm/vm_test.go
```diff
@@ -789,7 +789,7 @@ func TestStrptimeWithTimezone(t *testing.T) {
 	vm.t.Push("2012/01/18 06:25:00")
 	vm.t.Push("2006/01/02 15:04:05")
 	vm.execute(vm.t, obj.Program[0])
-	if vm.t.time != time.Date(2012, 01, 18, 06, 25, 00, 00, loc) {
+	if vm.t.time != time.Date(2012, 0o1, 18, 0o6, 25, 0o0, 0o0, loc) {
```
jaqx0r, line 792: Also bugs, just plain decimals here please.

internal/runtime/vm/vm_test.go
```diff
@@ -802,7 +802,7 @@ func TestStrptimeWithoutTimezone(t *testing.T) {
 	vm.t.Push("2012/01/18 06:25:00")
 	vm.t.Push("2006/01/02 15:04:05")
 	vm.execute(vm.t, obj.Program[0])
-	if vm.t.time != time.Date(2012, 01, 18, 06, 25, 00, 00, time.UTC) {
+	if vm.t.time != time.Date(2012, 0o1, 18, 0o6, 25, 0o0, 0o0, time.UTC) {
```
jaqx0r, line 805: and here.

jaqx0r (commented):
I'm sorry about the octal changes.  They're fine, except that it's highlighted some places where they should not have been interpreted as octal.

aRestless:
> Ok, so in Windows, a process's log library would handle copying its own log file, and then doing the truncation. Or is it an external process that would do this?
> 
> When you say you've seen this practice, can you give me an example of an application that does this?

An equivalent to "logrotate" as a standalone tool is very rare on Windows. One example of a reimplementation of the logrotate API for Windows can be found [here](https://github.com/plecos/logrotatewin). Logrotate does support the "copytruncate" mode, which is the one I was describing. But again, this is rarely used, and logs on Windows are often an afterthought.

My use case at the moment is Unreal Engine (and with that, all games built on it), which does the copy-then-truncate method on startup.

> One of the tests you edited is called examples_integration_test, which you've made a +build linux only test. If you can replicate the file-read part, the rest should be identical between linux and windows versions. Happy to see it in a followup.

That's a mistake then. All platform-specific tests should have received a proper file name suffix (sadly it also needs the build tag in the file, because _test and _platform suffixes don't work both on one file). Will rectify that and check if the test can be run on Windows.

internal/mtail/basic_tail_integration_test.go
aRestless: I think this ended up in the other file on accident. Will be reverted on the next push.

internal/mtail/examples_integration_unix_test.go
aRestless: done

internal/runtime/vm/vm_test.go
aRestless: done

internal/runtime/vm/vm_test.go
aRestless: done

internal/mtail/golden/reader_test.go
aRestless: done

aRestless:
I think the only concern left is the end to end tests. `TestExamplePrograms` is included, but that tests oneshot, not actually streamed data. Would you say there is value in an additional test that copies the content of the source file into the log file that is being read?

jaqx0r:
> I think the only concern left is the end to end tests. `TestExamplePrograms` is included, but that tests oneshot, not actually streamed data. Would you say there is value in an additional test that copies the content of the source file into the log file that is being read?

Ultimately yes I do.  But, that's perhaps a PR for another day?

jaqx0r (approved):

jaqx0r:
Looks like the linter is unhappy again, please fmt again with your rebase, thanks!

aRestless:
Fixed the linting errors, rebased and squashed. I hope everything is alright now, I had some issues with git on Windows (file endings and all that).

For the additional test I've got something around 90% finished and would likely need your input, but as you said that's something for a different PR.

jaqx0r (approved):

jaqx0r:
Thanks for working on this!

PR #648, opened 2022-05-28
Stabilize order of labels in statsd/graphite/collectd export

x-way:
When parsing statsd/graphite metrics with the positional [telegraf templates](https://docs.influxdata.com/telegraf/v1.22/data_formats/template-patterns/), the order of labels is critical.

Thus let's ensure that the order of labels in the export is stable by sorting them.

Signed-off-by: Andreas Jaggi <andreas.jaggi@waterwave.ch>

### DIFF
--- a/internal/exporter/export.go
+++ b/internal/exporter/export.go
@@ -13,6 +13,7 @@ import (
 	"io"
 	"net"
 	"os"
+	"sort"
 	"strings"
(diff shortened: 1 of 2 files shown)


jaqx0r (approved):

jaqx0r:
Thanks for the PR!

PR #667, opened 2022-07-29
Add http_debugging_endpoint and http_info_endpoint flag to allow disabeling http debug endpoints

robert-heinzmann-logmein:
fixes https://github.com/google/mtail/issues/305

Note: this is just https://github.com/google/mtail/pull/308 with the feedback incorporated (as the original went stale)

### DIFF
--- a/internal/mtail/httpstatus.go
+++ b/internal/mtail/httpstatus.go
@@ -23,6 +23,7 @@ const statusTemplate = `
 <p>Info: {{ if .HTTPInfoEndpoints }}<a href="/varz">varz</a>, <a href="/progz">progz</a> <a href="/tracez">tracez</a></p>{{ else }} disabled {{ end }}</p>
 <p>Debug: {{ if .HTTPDebugEndpoints }}<a href="/debug/pprof">debug/pprof</a>, <a href="/debug/vars">debug/vars</a>{{ else }} disabled {{ end }}</p>
 `
+
 const statusTemplateEnd = `
 </body>
 </html>
--- a/internal/mtail/httpstatus.go
+++ b/internal/mtail/httpstatus.go
@@ -11,6 +11,7 @@ import (
 )
 
 const statusTemplate = `
+<!DOCTYPE html>
 <html>
 <head>
 <title>mtail on {{.BindAddress}}</title>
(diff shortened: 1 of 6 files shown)


internal/mtail/mtail.go
```diff
@@ -96,19 +98,23 @@ func (m *Server) initHTTPServer() error {
 	}
 
 	mux := http.NewServeMux()
-	mux.HandleFunc("/favicon.ico", FaviconHandler)
-	mux.Handle("/", m)
-	mux.Handle("/progz", http.HandlerFunc(m.r.ProgzHandler))
-	mux.HandleFunc("/json", http.HandlerFunc(m.e.HandleJSON))
+	if m.httpDebugEndpoints {
+		mux.HandleFunc("/favicon.ico", FaviconHandler)
+		mux.Handle("/", m)
```
jaqx0r, line 103: These two previous are Info endpoints

robert-heinzmann-logmein: Moved this to the info endpoint section. Question would be, Should we render a different template in case debugging is disabled (internal/mtail/httpstatus.go) as the debug links would be dead then ?

robert-heinzmann-logmein:
Could anyone review this ?

jaqx0r:
Yep sorry.

internal/mtail/mtail.go
jaqx0r: I think we can pass the boolean into the template itself and only render those links if the boolean is set, right?

https://github.com/google/mtail/blob/main/internal/mtail/httpstatus.go#L36 copy `m.httpDebugEndpoints` into this struct and then update the `statusTemplate` to use `{{ if .httpDebugEndpoints }}` per https://pkg.go.dev/text/template

robert-heinzmann-logmein:
I incorporated the feedback.

jaqx0r (dismissed):

internal/mtail/httpstatus.go
```diff
@@ -11,15 +11,21 @@ import (
...
 <body>
 <h1>mtail on {{.BindAddress}}</h1>
 <p>Build: {{.BuildInfo}}</p>
-<p>Metrics: <a href="/json">json</a>, <a href="/graphite">graphite</a>, <a href="/metrics">prometheus</a>, <a href="/varz">varz</a></p>
-<p>Debug: <a href="/debug/pprof">debug/pprof</a>, <a href="/debug/vars">debug/vars</a>, <a href="/tracez">tracez</a>, <a href="/progz">progz</a></p>
+<p>Metrics: <a href="/json">json</a>, <a href="/graphite">graphite</a>, <a href="/metrics">prometheus</a></p>
+<p>Info: {{ if .HTTPInfoEndpoints }}<a href="/varz">varz</a>, <a href="/progz">progz</a> <a href="/tracez">tracez</a></p>{{ else }} disabled {{ end }}</p>
+<p>Debug: {{ if .HTTPDebugEndpoints }}<a href="/debug/pprof">debug/pprof</a>, <a href="/debug/vars">debug/vars</a>{{ else }} disabled {{ end }}</p>
+`
+const statusTemplateEnd = `
```
jaqx0r, line 26: Why does this need to be a separate template?

robert-heinzmann-logmein: It just closes the HTML to be "properly" formatted. As we have conditional rendering of the status and runtime stats tables (depending on command line flags), adding this closing to the status and runtime rendering looks more complex to me. So this looks like the easiest path for me.

robert-heinzmann-logmein: Fixed the linter error.

jaqx0r: Ah yep I see why it needs to be like this now.  Thanks!

jaqx0r (approved):

robert-heinzmann-logmein:
Sorry, but I do not know how to fix the remaining CI issues

jaqx0r:
Thanks!

PR #676, opened 2022-08-17
Fix memory leak on corrupt unicode character

rreuvekamp:
Fix for a memory leak which is hitting us on v3.0.0-rc50.

When a corrupt unicode character is received, mtail stops processing logs and instead buffers them in memory. It happens because a recent change ( c12a012 ) has made it so parsing of bytes in the log buffer to runes is being retried when parsing results in a RuneError. I guess that change was made under the assumption that rune errors are always recoverable (you just need to increase the buffer), but that is not the case. In production we've seen this memory leak happen multiple times since we deployed the new version of mtail.

With this commit RuneErrors on parsing unicode characters are ignored when there are enough characters after it in the buffer.
This keeps the functionality intact that bytes that gave a RuneError are being reparsed, but with a larger buffer, to not break unicode characters which were split in half from the buffered reading on the logfile being read.

### DIFF
--- a/internal/tailer/logstream/decode.go
+++ b/internal/tailer/logstream/decode.go
@@ -21,15 +21,23 @@ func decodeAndSend(ctx context.Context, lines chan<- *logline.LogLine, pathname 
 	var (
 		r     rune
 		width int
+		count int
 	)
 	var i int
 	for ; i < len(b) && i < n; i += width {
 		r, width = utf8.DecodeRune(b[i:])
 		if r == utf8.RuneError {
 			if len(b)-i > 10 {
+				// If there are more than enough bytes in the buffer
+				// after this, ignore the error as it's not fixable.
+				// If not, return so that the caller can try again
+				// with a larger buffer, in the event that a unicode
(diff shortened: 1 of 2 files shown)


jaqx0r (approved):
Thanks, this is great!

jaqx0r:
https://github.com/google/mtail/runs/7910805252?check_suite_focus=true has a failure in `make fuzz-regtest`.

There are debugging instructions at https://google.github.io/mtail/debugging.html#fuzzer-crashes

rreuvekamp:
@jaqx0r thank you for the review.

I cannot reproduce the fuzzer fail locally. It seems CI timeouts on a certain test. Could it be that this is a problem in CI and not so much in my patch?

jaqx0r:
Yeah I think so.  I've disabled that from the branch requirements; Can you update your branch and then I can merge it.

rreuvekamp:
Good to know. I've updated the branch

jaqx0r (approved):

jaqx0r:
Thanks!

PR #700, opened 2022-12-22
Update Go modules

SuperQ:
Update several Go modules that have timestamp/commit versions that are not handled by dependabot.
* Bump some indirect deps to satisfy go mod tidy version handling.

Signed-off-by: SuperQ <superq@gmail.com>


jaqx0r (approved):
Thanks!

jaqx0r:
As soon as I've fixed the problem with goreleaser I'll merge.

PR #767, opened 2023-09-15
examples/postfix.mtail - move dash to end of character class

andreby:
Otherwise the "application" regex group won't match a string such as "postfix/postfix-script"


jaqx0r (approved):

jaqx0r:
Thanks :)

Issue #675, opened 2022-08-10
Histogram lower bucket and "0" observations expectations


filippog:
Hi,
thank you for mtail, I'm following up from #226 where it was reported mtail histogram buckets don't match Prometheus' specs/expectations, specifically with regard to `0` observations.

I can confirm that the issue is still present in rc50, IOW observing a 0 value increments `_count` but not the lowest bucket:

```
$ cat issue-226.mtail 
histogram foo_requests buckets -1, 2, 4, 6 by foo

/^(?P<foo>[a-z]+) (?P<time>\d+)$/ {
  foo_requests[$foo] = $time
}
```

```
$ ./mtail -progs issue-226.mtail -logs /tmp/foo.log &
$ echo "c 0" >> /tmp/foo.log
```

```
$ curl -s localhost:3903/metrics | grep -i requests
# HELP foo_requests defined at issue-226.mtail:1:11-22
# TYPE foo_requests histogram
foo_requests_bucket{foo="c",prog="issue-226.mtail",le="2"} 0
foo_requests_bucket{foo="c",prog="issue-226.mtail",le="4"} 0
foo_requests_bucket{foo="c",prog="issue-226.mtail",le="6"} 0
foo_requests_bucket{foo="c",prog="issue-226.mtail",le="+Inf"} 0
foo_requests_sum{foo="c",prog="issue-226.mtail"} 0
foo_requests_count{foo="c",prog="issue-226.mtail"} 1
```

This caught us by surprise when using mtail's native histograms and exporting to Prometheus. The "add a `-1` bucket" workaround does indeed work though I think histograms should match what users expect (i.e. all observations are <= their buckets). What do you think? Alternatively the issue could be documented (and perhaps linted for?)

jaqx0r:
Sounds like a bug; the name of the bucket is `le` meaning less-than-or-equal, so the lower bound intentionally goes into the bucket above, but zero is indeed a special case here.

RichieSams:
Is there an ETA on a fix for this? I'm also hitting this issue

havlicekj:
This should now be fixed by https://github.com/google/mtail/pull/803

Issue #226, opened 2019-04-09
First custom bucket in histogram metric not created correctly


biz812:
Hi,

I've noticed there a problem with the first bucket in a custom histogram, in that it doesn't appear to be created correctly or, possibly, at all.

I am using:
`mtail_build_info{branch="HEAD",goversion="go1.11.6",revision="b5aa95d10fb3af7b843e3814bc65aedd983425d9",version="v3.0.0-rc29"} 1`

I am running mtail as follows:
`/usr/local/sbin/mtail -progs /tmp/mtail/foo.mtail -logs /tmp/foo.log`

You can reproduce using this simple program:
```
histogram foo_requests buckets 2, 4, 6 by foo

/^(?P<foo>[a-z]+) (?P<time>\d+)$/ {
  foo_requests[$foo] = $time
}
```

When you create a log entry with a value greater than the first bucket you get metrics:
```
# echo "b 3" >> /tmp/foo.log

# curl -s localhost:3903/metrics|grep request
# HELP foo_requests defined at foo.mtail:1:11-22
# TYPE foo_requests histogram
foo_requests_bucket{foo="b",prog="foo.mtail",le="4"} 1
foo_requests_bucket{foo="b",prog="foo.mtail",le="6"} 1
foo_requests_bucket{foo="b",prog="foo.mtail",le="+Inf"} 1
foo_requests_sum{foo="b",prog="foo.mtail"} 3
foo_requests_count{foo="b",prog="foo.mtail"} 1
```
But you'll notice that the `le="2"` is not there and set to `0` as you would expect.

When you make a request that falls into the first bucket (`0-2`):
```
# echo "c 1" >> /tmp/foo.log

# curl -s localhost:3903/metrics|grep request
# HELP foo_requests defined at foo.mtail:1:11-22
# TYPE foo_requests histogram
foo_requests_bucket{foo="b",prog="foo.mtail",le="4"} 1
foo_requests_bucket{foo="b",prog="foo.mtail",le="6"} 1
foo_requests_bucket{foo="b",prog="foo.mtail",le="+Inf"} 1
foo_requests_sum{foo="b",prog="foo.mtail"} 3
foo_requests_count{foo="b",prog="foo.mtail"} 1
foo_requests_bucket{foo="c",prog="foo.mtail",le="4"} 0
foo_requests_bucket{foo="c",prog="foo.mtail",le="6"} 0
foo_requests_bucket{foo="c",prog="foo.mtail",le="+Inf"} 0
[...]

PR #803, opened 2024-02-05
Histogram zero fix

havlicekj:
Currently, mtail checks both lower and upper bound when searching for a histogram bucket. This is not only unnecessary (because buckets are constructed so that they are sorted and continuous), it leads to bugs like https://github.com/google/mtail/issues/226 (fixed in the past), https://github.com/google/mtail/issues/675 (still open) and the fact that negative values will not be counted in any bucket unless you explicitly add negative bucket that is lower than any value that will be observed (which is not a problem for timings, but histograms could be used for other things).

Behavior of other prometheus clients is to look for first bucket where observed value less or equal to its upper bound. This MR makes mtail to also behave this way, which ensures any observation will always be counted in some bucket. I also changed integration test to check observations of zero are counted and simplified the histogram bucket initialization a bit.

### DIFF
--- a/internal/metrics/datum/buckets.go
+++ b/internal/metrics/datum/buckets.go
@@ -44,7 +44,7 @@ func (d *Buckets) Observe(v float64, ts time.Time) {
 	defer d.Unlock()
 
 	for i, b := range d.Buckets {
-		if b.Range.Contains(v) {
+		if v <= b.Range.Max {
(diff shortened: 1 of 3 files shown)


jaqx0r:
Thanks for the change and the detailed explanation.

jaqx0r (approved):

Issue #685, opened 2022-10-10
Reading from `/dev/stdin` is very slow


mpol:
Processing this 7.5M line file normally takes ~2-3 minutes on my laptop:

```
$ time ./mtail -emit_prog_label=false -one_shot -one_shot_format=prometheus -logtostderr -progs=lc -logs access-ssl.log.6
I1010 12:23:55.906123  179121 main.go:117] mtail version 3.0.0-rc50 git revision 89b462d7a83bab283e84a3b6e592a964090f7f4c go version go1.18.3 go arch amd64 go os linux
I1010 12:23:55.906203  179121 main.go:118] Commandline: ["./mtail" "-emit_prog_label=false" "-one_shot" "-one_shot_format=prometheus" "-logtostderr" "-progs=lc" "-logs" "access-ssl.log.6"]
I1010 12:23:55.906590  179121 store.go:189] Starting metric store expiry loop every 1h0m0s
I1010 12:23:55.907672  179121 runtime.go:188] Loaded program lc.mtail
I1010 12:23:55.907708  179121 runtime.go:84] unmarking lc.mtail
I1010 12:23:55.907763  179121 logstream.go:61] Parsed url as /tmp/access-ssl.log.6
I1010 12:23:55.907871  179121 filestream.go:278] signalling stop at next EOF
I1010 12:23:55.907896  179121 tail.go:287] Tailing /tmp/access-ssl.log.6
I1010 12:23:55.907986  179121 tail.go:343] No polling loop in oneshot mode.
I1010 12:23:55.908012  179121 mtail.go:126] Listening on [::]:3903
I1010 12:23:55.908022  179121 tail.go:315] No gc loop in oneshot mode.
I1010 12:26:51.737588  179121 runtime.go:288] END OF LINE
I1010 12:26:51.737615  179121 vm.go:1059] VM "lc.mtail" finished
# HELP lines defined at lc.mtail:1:9-13
# TYPE lines counter
lines 7.519629e+06

real	2m55,853s
user	3m37,175s
sys	0m29,575s
```

But when it is piped on standard input like this:

```
$ cat access-ssl.log.6 | time ./mtail -emit_prog_label=false -one_shot -one_shot_format=prometheus -logtostderr -progs=lc -logs /dev/stdin 
```

it takes forever, reading ~2.5k lines/minute:

```
$ http :3903/metrics|grep stdin; sleep 5m;  http :3903/metrics|grep stdin
[...]

PR #878, opened 2024-06-10
Reduce wait friction when reading from a pipe

rideliner:
Uses logic similar to `fileStream` to circumvent waiting when there may be more to read immediately on a pipe.


jaqx0r:
Thanks for the idea!  This is the real fix for #685 .

I've got some pending changes to the logstreams so I've applied a similar fix in my next PR.

jaqx0r:
Unfortunately this PR contains a race condition in test by changing the arg to NewTest to zero.  Fixing this isn't trivial.

Issue #810, opened 2024-02-16
2 `subst()` calls within a single scope breaks compilation


EconomicTouristsArmLate:
Having 2 `subst()` calls within a single scope breaks with the error message:

```
Redeclaration of capture group `0' previously declared at apache2.access.mtail:63:20-24
```

This example works fine:
```
gauge foo by asdf
gauge bar by asdf
/(?P<asdf>.+)/ {
  foo[subst(/\d/, "", $asdf)]++
  /.+/ {
    bar[subst(/\d/, "", $asdf)]++
  }
}
```

while this throws the `Redeclaration of capture group 0` error:
```
gauge foo by asdf
gauge bar by asdf
/(?P<asdf>.+)/ {
  foo[subst(/\d/, "", $asdf)]++
  bar[subst(/\d/, "", $asdf)]++
}
```

previous report: https://github.com/google/mtail/issues/693

jaqx0r:
The problem is that the regular expression creates the numbered match
variables automatically, and the second regular expression then conflicts
in the same scope as the first.

There are two issues:

1) the numbered matches are not referenced so they should not matter

2) as the argument to subst the scope is even smaller and they should not
be leaking into the calling scope

Thanks for the report!

Issue #693, opened 2022-11-17
regex  ..Redeclaration of capture group `0' previously declared


li35619081:
filename=/20221117/com.test.api.ApiService/queryList/10.1.3.4/10.1.5.4/consumer.success

classname=com.test.api.ApiService
ip1=10.1.3.4
ip2=10.1.5.4
I want to get this information from the filename

==================
gauge filename1_success by method
gauge filename2_success by method

hidden text filename0
hidden text filename1
hidden text filename2

#/20221117/com.test.api.ApiService/queryList/10.1.3.4/10.1.5.4/consumer.success
/(?P<date>\d+) (?P<count>\d+)/ {
	filename0=getfilename()
	#20221117
	filename1=subst(/\/\d+/, "",filename0)   
	filename2=subst(/(([a-zA-Z_][a-zA-Z0-9_]*[.])*([a-zA-Z_][a-zA-Z0-9_]*))/, "",filename0)  
	filename1_success[filename1] += int($count)
	filename2_success[filename2] += int($count)
	 
}

================

`Redeclaration of capture group `0' previously declared at test.mtail:14:18-24`

jaqx0r:
Can you highlight what's at line 14?  I can't tell from your report.

EconomicTouristsArmLate:
Also ran into this just now. The issue here is the second `subst()`.
Apparently, having 2 `subst` calls within a scope breaks with the above error message.

This example works fine:
```
gauge foo by asdf
gauge bar by asdf
/(?P<asdf>.+)/ {
  foo[subst(/\d/, "", $asdf)]++
  /.+/ {
    bar[subst(/\d/, "", $asdf)]++
  }
}
```

while this throws the `Redeclaration of capture group 0` error:
```
gauge foo by asdf
gauge bar by asdf
/(?P<asdf>.+)/ {
  foo[subst(/\d/, "", $asdf)]++
  bar[subst(/\d/, "", $asdf)]++
}
```

PR #906, opened 2024-08-05
fix: fix symbol checking and creation for ``subst`` with a regexp

terencehonles:
This change updates the ``checkRegex`` helper to not track symbols for a regexp when it is used by the builtin ``subst``. In this case the VM will not store these matches, and therefore these symbols will not be available to the program. The compiler will, however, compile these regexp and add them to the regexp table for use by the VM, and multiple calls to ``subst`` are already handled by the VM.

fixes: https://github.com/google/mtail/issues/693, fixes: https://github.com/google/mtail/issues/810

### DIFF
--- a/internal/runtime/compiler/checker/checker.go
+++ b/internal/runtime/compiler/checker/checker.go
@@ -35,6 +35,7 @@ type checker struct {
 	tooDeep           bool
 	maxRecursionDepth int
 	maxRegexLength    int
+	noRegexSymbols    bool
 }
 
 // Check performs a semantic check of the astNode, and returns a potentially
@@ -81,6 +82,12 @@ func (c *checker) VisitBefore(node ast.Node) (ast.Visitor, ast.Node) {
 		n.Scope = symbol.NewScope(c.scope)
 		c.scope = n.Scope
 		glog.V(2).Infof("Created new scope %v in condstmt", n.Scope)
+		return c, n
+
+	case *ast.BuiltinExpr:
+		if n.Name == "subst" {
+			c.noRegexSymbols = true
+		}
 		return c, n
 
 	case *ast.CaprefTerm:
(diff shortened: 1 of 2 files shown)


internal/runtime/compiler/checker/checker.go
```diff
@@ -838,7 +849,9 @@ func (c *checker) VisitAfter(node ast.Node) ast.Node {
 			return n
 		}
 		n.Pattern = pe.pattern.String()
-		c.checkRegex(n.Pattern, n)
+		if !c.skipRegexCheck {
```
jaqx0r, line 852: `checkRegex` also does a parse test that is now being skipped for subst arguments, so this weakens the pass, deferring regex syntax errors until code generation time.

I'd prefer this be passed as an argument, to disable the `capNames` loop on checker.go:894.

I'd actually prefer a better way of handling symbol insertion such that positional capref names are only checked if they're referenced later on.  This might mean adding a per-regexp scope and then promoting referenced identifiers out of that scope, which will be weird and require special logic in the symbol lookup.

Or... allow the symbol table to contain mutiple symbols under a name, and then throw the error in `checkSymbolTable()` if there is more than one assigned to a name.

jaqx0r (changes requested):

internal/runtime/compiler/checker/checker.go
terencehonles: I see what you're saying and I've updated this and added tests to make sure this is still checked (and verified that before making the updates the tests were failing, which is what you were pointing out).

Cross referencing the grammar and the VM it looks like the only way `t.matches` is updated is if a `code.Match` or `code.Smatch` opcode is emitted which appears to be a `/pattern/ {}` or a `var [=!]~ /pattern/`, so it doesn't look like these symbols should even exist on the checker's symbol table.

I think the two other symbol table handling changes can still apply as further improvements to the code, but I'm not sure how often they occur since the two opcodes I listed seem like they would be creating new scopes and I'd have to understand the VM and how variables are looked up to really update the checker to match the VM.

jaqx0r: Yep I agree.  I had a think about this last night after I wrote the comment.  The meaning of the regular expression here is not one that implies variable references.  At best one might think that other parameters to subst could be reference names but there is no scope leaked outside the function parameters.

I will take the task of fixing the symbol table code.

jaqx0r (approved):

Issue #929, opened 2024-10-17
This repo is obsolete, please follow `jaqx0r/mtail` instead


jaqx0r:
No `mtail` maintainers have write access to this repository anymore so please repoint your git remote upstreams to `https://github.com/jaqx0r/mtail` and docker images to `ghcr.io/jaqx0r/mtail`.  Documentation now hosted at https://jaqx0r.github.io/mtail

avesst:
@jaqx0r I think more information is warranted here. Could you elaborate on what happened to this repository?

anarcat:
it would be great to have some official word from @google / @google-admin that this is abandoned? perhaps archiving the repo and pointing to the new one is in order?

anarcat:
> It's not abandoned. Work continues in jaqx0r/mtail as already stated.

I understand, but you might understand why people could be a little
hesitant into switching over from an official-looking @google repository
to what seems at first glance like a third-party fork.

I understand you're the primary maintainer here, but it would still be
good, IMHO, to archive the @google repository and point to yours to make
things official.

czerasz:
A mention in `README.md` would be nice! IMHO this is the minimum...

Otherwise how does someone know, that [@google](https://github.com/google) repository is not active and work is continued on https://github.com/jaqx0r/mtail?

jaqx0r:
@chenchuanpeng Yes please.

link2xt:
There is also an announcement here: https://groups.google.com/g/mtail-users/c/BpYIHnhxLXM

PR #940, opened 2024-12-26
Add support for Apache Kafka log source

wanmail:
### Why use Apache Kafka
In our environment, it is difficult for me to deploy a local log analysis tool directly on the machine. Fortunately, they have collected the logs into Kafka. I only need to consume them in Kafka.

### About Unit Test
I wrote a simple unit test with reference to TestFileStreamRead. But I started a Kafka container locally before running the unit test, so the unit test will fail if directly `make check test`.

But I have been running it in our production environment for a week, and process about 100 million Kafka messages every day.

### About not using waker
Each message in Kafka is a log line, there is no need to use LineReader to find line endings, so I did not use waker.Waker.

Only handled `ctx done` and `network connection error`.


jaqx0r:
Development on this fork has ceased.  See #929

PR #985, opened 2025-09-15
FEAT: allow longstream.New pathname with hash

nmreadelf:


jaqx0r:
Please create this PR against jaqx0r/mtail per #929

PR #993, opened 2025-11-03
Unknown reader errors should close reader/tailer to avoid mtail looping on broken fds.

vaLski:
If mtail is tailing a file under a mount, and this mount is gone, mtail reader might block on

	read error: bad file descriptor ....

This will DoS mtail which won't respond to standard signals and significantly increase its CPU consumption.

Better strategy is to simply close reader/tailer on the broken fd and let the tailer do its job.

Another retryable cases like syscall.ESTALE can be added in future, but for now fail fast seems like a logical approach.


jaqx0r:
Please create this PR against jaqx0r/mtail per #929

